PR #13, opened 2023-10-26
Add support for using a connection url so we can specify connection p…

langchain-infra:
…arams

### DIFF
--- a/charts/langsmith/templates/playground/deployment.yaml
+++ b/charts/langsmith/templates/playground/deployment.yaml
@@ -18,7 +18,7 @@ spec:
   {{- end }}
   selector:
     matchLabels:
-      {{- include "langsmith.labels" . | nindent 6 }}
+      {{- include "langsmith.selectorLabels" . | nindent 6 }}
       app.kubernetes.io/name: {{ include "langsmith.fullname" . }}-{{ .Values.playground.name }}
   template:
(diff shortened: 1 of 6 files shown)


charts/langsmith/templates/playground/deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
   {{- end }}
   selector:
     matchLabels:
-      {{- include "langsmith.labels" . | nindent 6 }}
+      {{- include "langsmith.selectorLabels" . | nindent 6 }}
```
langchain-infra, line 21: needed otherwise version bumps cause problems

charts/langsmith/values.yaml
```diff
@@ -14,27 +14,27 @@ images:
...
-    pullPolicy: Always
-    tag: "latest"
+    pullPolicy: IfNotPresent
+    tag: "99d8f59"
   playgroundImage:
     repository: "docker.io/langchain/langchainplus-playground"
-    pullPolicy: Always
-    tag: "latest"
+    pullPolicy: IfNotPresent
+    tag: "99d8f59"
```
langchain-infra, line 30: use static tags so that people don't accidentally pull in changes they aren't ready for

agola11 (approved):

PR #19, opened 2023-12-07
Fix queue environment variables and update chart images

langchain-infra:


jakerachleff (approved):
I don't know what this means but ship it

PR #46, opened 2024-01-18
Add migration docs

langchain-infra:


charts/langsmith/docs/UPGRADE-0.2.x.md
```diff
@@ -0,0 +1,118 @@
...
+    [{"id":"00000000-0000-0000-0000-000000000000","has_waitlist_access":true,"created_at":"2023-09-13T18:25:10.488407","display_name":"Personal","config":{"is_personal":true,"max_identities":1},"tenant_handle":"default"}]%
+    ```
+
+3. Visit the external ip for the `langsmith-frontend` service on your browser
+
+    The Langsmith UI should be visible/operational
+
+    ![.langsmith_ui.png](../langsmith_ui.png)
+
+Note that old runs will not be visibile in the UI until you migrate them. See [Migrating old runs](#migrating-old-runs) for more information.
```
barberscott, line 63: Replace `until` with `unless` and add something like:  "This step is only required if you need access to run data from before the upgrade to Langsmith 0.2.x.  If you are unsure or need assistance, please reach out to support@langchain.dev and we'd be happy to help."

charts/langsmith/README.md
```diff
@@ -184,7 +185,7 @@ We typically validate deployment using the following Jupyter notebook:
 ## FAQ:
 
 1. How can we upgrade our application.
```
barberscott, line 187: Nit:  Should end in ?

barberscott (approved):
One minor nit, otherwise looks good.

PR #52, opened 2024-01-23
Migrate no auth

langchain-infra:


charts/langsmith/scripts/migrate_no_auth.sh
```diff
@@ -0,0 +1,24 @@
...
+postgres_url=$1
+admin_user_email=$2
+
+#PostgreSQL query
+update_query_string=$(cat <<EOF
+WITH admin_user_id AS (
+    SELECT id FROM public.users WHERE email = '$admin_user_email'
+)
+INSERT INTO public.identities(id, tenant_id, user_id, created_at, read_only)
+VALUES ('00000000-0000-0000-0000-000000000000', '00000000-0000-0000-0000-000000000000', (SELECT id FROM admin_user_id), NOW(), false);
```
nfcampos, line 19: Does this create a personal or shared tenant?

charts/langsmith/docs/MIGRATE-OAUTH.md
```diff
@@ -0,0 +1,40 @@
+# Migrating from no-auth to OAuth
+
+We have provided a migration script to migrate your no-auth tenant to an oauth tenant. The script is located in the `scripts` directory of this repository.
+
+### Prerequisites
+
+Ensure you have the following tools/items ready.
```
nfcampos, line 7: This should say the script needs to run after they add oauth and login to the UI with the desired "admin user" at least once

charts/langsmith/scripts/migrate_no_auth.sh
langchain-infra: I'm just creating an identity in the tenant here. Not actually creating a tenant, just making the  default one shared and adding a user to it

charts/langsmith/docs/MIGRATE-OAUTH.md
langchain-infra: fixed

charts/langsmith/scripts/migrate_no_auth.sh
nfcampos: Hmm yes that's right, I believe that singleton tenant is a personal tenant by default. This would then leave this user in a funny state with two personal tenants, so I think might be best to also edit update this tenant's config in this script to mark it as shared?

langchain-infra: I don't think we actually create a singleton user though

langchain-infra: omg i just realized i didnt copy the ful sql query lol good catch

langchain-infra: fixed

nfcampos (approved):

PR #66, opened 2024-02-05
Infra/add extra continer config argument

langchain-infra:


charts/langsmith/values.yaml
```diff
@@ -77,6 +77,7 @@ backend:
 #        cpu: 200m
 #        memory: 500Mi
     command: []
+    extraContainerConfig: {}
```
nfcampos, line 80: is this a breaking change for helm chart users?

langchain-infra: nope, none of these are breaking just allow specifying some additional things

PR #72, opened 2024-02-23
Infra/update nginx size

langchain-infra:
All requests go through nginx so we should bump the client body size.


charts/langsmith/templates/frontend/config-map.yaml
```diff
@@ -81,7 +83,9 @@ data:
   nginx.conf: |
     server {
       listen       {{ .Values.frontend.containerPort }};
+      listen  [::]:{{ .Values.frontend.containerPort }};
```
langchain-infra, line 86: allows us to listen to ipv6 as well

nfcampos (approved):

PR #74, opened 2024-03-02
feat: Add oai cache to our frontend container

langchain-infra:
Ad the option to enable an oai cache on your langsmith instance.

### DIFF
--- a/charts/langsmith/values.yaml
+++ b/charts/langsmith/values.yaml
@@ -219,6 +219,9 @@ frontend:
   name: "frontend"
   containerPort: 8080
   existingConfigMapName: ""
+  # Turn on LLM Proxy Cache
+  cache:
+    enabled: true
   deployment:
     replicas: 1
     labels: {}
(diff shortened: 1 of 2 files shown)


charts/langsmith/values.yaml
```diff
@@ -219,6 +219,9 @@ frontend:
   name: "frontend"
   containerPort: 8080
   existingConfigMapName: ""
+  # Turn on LLM Proxy Cache
+  cache:
```
langchain-infra, line 223: should this be on or off?

agola11: Probably default on? People need to explicitly change oai url to use anyways?

agola11 (approved):

charts/langsmith/values.yaml
langchain-infra: also any configuration params we should add here? I can also add some otehr ones like anthropic

PR #76, opened 2024-03-08
Infra/bump charts fixes

langchain-infra:
Bump to 0.1.18 and add a bunch of examples. Swap to SAQ.


charts/langsmith/values.yaml
```diff
@@ -219,11 +229,6 @@ frontend:
   name: "frontend"
   containerPort: 8080
   existingConfigMapName: ""
-  # Turn on LLM Proxy Cache
-  cache:
```
langchain-infra, line 223: we deploy via a separate image now, will add in separate pr

PR #77, opened 2024-03-09
Handle cors and cache control in nginx unit

nfcampos:


charts/langsmith/templates/backend/config-map.yaml
```diff
@@ -11,9 +11,38 @@ data:
     {
       "listeners": {
           "*:{{ .Values.backend.containerPort }}": {
-              "pass": "applications/fastapi"
+              "pass": "routes/main"
```
langchain-infra, line 14: also need to bump chart.yaml

langchain-infra (approved):

charts/langsmith/Chart.yaml
```diff
@@ -6,4 +6,4 @@ maintainers:
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
 version: 0.3.6
```
langchain-infra, line 8: sorry should be this version

PR #90, opened 2024-04-17
move queues to configmap

langchain-infra:


langchain-infra:
> This is the one for our infra, we can leave the default for self hosted if that's what this is doing
> 
> ```
>     INGESTION_QUEUE: str = "default"
>     ADHOC_QUEUE: str = "adhoc"
>     RUN_RULES_QUEUE: str = "rules"
> ```

I know i need this because we use the self-hosted chart and we can't duplicate environment variables

jakerachleff (approved):

PR #92, opened 2024-04-25
feat(auth): support basic auth

bvs-langchain:
Adds configuration required to support [basic auth mode](http://localhost:3000/reference/authentication_authorization/authentication_methods#self-hosted) along with example configuration values


charts/langsmith/templates/backend/auth-bootstrap.yaml
```diff
@@ -0,0 +1,96 @@
...
+      serviceAccountName: {{ include "backend.serviceAccountName" . }}
+      containers:
+        - name: auth-bootstrap
+          {{- with .Values.backend.authBootstrap.command }}
+          command:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          env:
+            {{- include "langsmith.commonEnv" . | nindent 12 }}
+            - name: BASIC_AUTH_ENABLED
```
langchain-infra, line 49: should this be in our commonEnv or envConfigmap(they go in configmap if we want to override in our saas app)

charts/langsmith/values.yaml
```diff
@@ -135,6 +144,30 @@ backend:
...
+    resources: {}
+    #      limits:
+    #        cpu: 1000m
+    #        memory: 1Gi
+    #      requests:
+    #        cpu: 200m
+    #        memory: 500Mi
+    command:
+      - "python"
+      - "hooks/auth_bootstrap.pyc"
```
langchain-infra, line 162: interesting, til we could use the .pyc instead of .py for these scripts

charts/langsmith/templates/backend/auth-bootstrap.yaml
bvs-langchain: I think commonEnv? won't it be by default `false` for the SaaS version based on `values.yaml`, so no need to add to configmap?

charts/langsmith/templates/backend/secrets.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if not .Values.config.basicAuth.existingSecretsName }}
+
+{{- $secretName := include "langsmith.basicAuthSecretsName" . }}
+{{- $secretObj := (lookup "v1" "Secret" .Release.Namespace $secretName) | default dict }}
+{{- $secretData := (get $secretObj "data") | default dict }}
+
+{{- $jwtSecretFromValues := .Values.config.basicAuth.jwtSecret | default "" }}
+{{- $jwtSecretFromSecret := (get $secretData "jwt_secret") | default "" }}
+{{- $jwtSecret := ternary $jwtSecretFromSecret (b64enc $jwtSecretFromValues)  (empty $jwtSecretFromValues) }}
```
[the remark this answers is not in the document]
bvs-langchain: I was actually thinking we should autogenerate if values are empty and no existing secret name is given - wdyt? https://blog.cloudcover.ch/posts/reusing-existing-kubernetes-secrets-in-helm-templates/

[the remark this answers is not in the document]
langchain-infra: So i actually did this for a previous chart but the issue is this fails with argo/helm template(for people who don't use helm cli directly) since the lookup will always fail

charts/langsmith/templates/backend/auth-bootstrap.yaml
langchain-infra: agreed lets put in commonEnv then(does seem like its used in multiple places)

charts/langsmith/templates/backend/secrets.yaml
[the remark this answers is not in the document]
bvs-langchain: Ok I'll update 👍

charts/langsmith/templates/_helpers.tpl
```diff
@@ -199,6 +200,19 @@ Template containing common environment variables that are used by several servic
...
       optional: true
+- name: BASIC_AUTH_ENABLED
+  value: {{ .Values.config.basicAuth.enabled | quote }}
+{{- if .Values.config.basicAuth.enabled }}
+- name: BASIC_AUTH_JWT_SECRET
+  valueFrom:
+    secretKeyRef:
+      name: {{ include "langsmith.secretsName" . }}
+      key: jwt_secret
+- name: FF_ORG_CREATION_DISABLED
```
langchain-infra, line 211: these can be configured separately, should we just use those config values?

langchain-infra: or at least org_creation_disabled

charts/langsmith/templates/backend/auth-bootstrap.yaml
```diff
@@ -0,0 +1,94 @@
...
+metadata:
+  name: {{ include "langsmith.fullname" . }}-{{ .Values.backend.name }}-basic-auth-bootstrap
+  labels:
+    {{- include "langsmith.labels" . | nindent 4 }}
+    {{- with.Values.backend.authBootstrap.labels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-weight": "0"
```
langchain-infra, line 13: should we also add an argocd postsync annotation(so triggered in argo as well)

charts/langsmith/templates/_helpers.tpl
```diff
@@ -250,6 +264,7 @@ Template containing common environment variables that are used by several servic
 {{- end }}
 {{- end }}
 
+
```
langchain-infra, line 267: intended?

langchain-infra (approved):

PR #102, opened 2024-06-20
Add options for TTL

akira:
```
config:
  ttl:
    enabled: true
    upgrade_enabled: true
    ttl_period_seconds:
      longlived: <ll_seconds>
      shortlived: <sl_seconds>
```

To set:
```
 - name: FF_TRACE_TIERS_ENABLED
    value: "true"
  - name: FF_UPGRADE_TRACE_TIER_ENABLED
    value: "true"
  - name: TRACE_TIER_TTL_DURATION_SEC_MAP
    value: "{ \"longlived\": 1234, \"shortlived\": 567 }"
```

### DIFF
--- a/charts/langsmith/templates/_helpers.tpl
+++ b/charts/langsmith/templates/_helpers.tpl
@@ -200,19 +200,13 @@ Template containing common environment variables that are used by several servic
       key: openai_api_key
 - name: GO_ENDPOINT
   value: http://{{- include "langsmith.fullname" . }}-{{.Values.platformBackend.name}}:{{ .Values.platformBackend.service.port }}
-{{- if .Values.config.ttl }}
-{{- if .Values.config.ttl.enabled }}
 - name: FF_TRACE_TIERS_ENABLED
-  value: true
-{{- if .Values.config.ttl.upgrade_enabled }}
+  value: {{ .Values.config.ttl.enabled }}
 - name: FF_UPGRADE_TRACE_TIER_ENABLED
-  value: true
-{{- end }}
-{{- if .Values.config.ttl.ttl_period_seconds }}
+  value: {{ .Values.config.ttl.upgrade_enabled }}
+{{- if .Values.config.ttl.enabled }}
 - name: TRACE_TIER_TTL_DURATION_SEC_MAP
   value: "{ \"longlived\": {{ .Values.config.ttl.ttl_period_seconds.longlived }}, \"shortlived\": {{ .Values.config.ttl.ttl_period_seconds.shortlived }} }"
-{{- end }}
-{{- end }}
 {{- end }}
 - name: FF_ORG_CREATION_DISABLED
   value: {{ .Values.config.orgCreationDisabled | quote }}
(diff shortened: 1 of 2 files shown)


charts/langsmith/templates/_helpers.tpl
```diff
@@ -200,6 +200,20 @@ Template containing common environment variables that are used by several servic
       key: openai_api_key
 - name: GO_ENDPOINT
   value: http://{{- include "langsmith.fullname" . }}-{{.Values.platformBackend.name}}:{{ .Values.platformBackend.service.port }}
+{{- if .Values.config.ttl }}
```
langchain-infra, line 203: since there are default values, can we remove the ifs?

charts/langsmith/values.yaml
```diff
@@ -72,6 +72,13 @@ config:
     enabled: false
     oauthClientId: ""
     oauthIssuerUrl: ""
+  ttl:
+    enabled: false
+    upgrade_enabled: false
```
langchain-infra, line 77: could we explain upgrades in a comment?

charts/langsmith/templates/_helpers.tpl
akira: Yeah - maybe leave the if for the duration map?  Also, how about backwards compability?  If they don't have these it will fail without the if

charts/langsmith/values.yaml
akira: Updated comment

langchain-infra (approved):

charts/langsmith/templates/_helpers.tpl
akira: OK just left it for the duration map

charts/langsmith/templates/_helpers.tpl
```diff
@@ -200,6 +200,14 @@ Template containing common environment variables that are used by several servic
       key: openai_api_key
 - name: GO_ENDPOINT
   value: http://{{- include "langsmith.fullname" . }}-{{.Values.platformBackend.name}}:{{ .Values.platformBackend.service.port }}
+- name: FF_TRACE_TIERS_ENABLED
+  value: {{ .Values.config.ttl.enabled }}
```
langchain-infra, line 204: need to | quote this

charts/langsmith/templates/_helpers.tpl
```diff
@@ -201,9 +201,9 @@ Template containing common environment variables that are used by several servic
 - name: GO_ENDPOINT
   value: http://{{- include "langsmith.fullname" . }}-{{.Values.platformBackend.name}}:{{ .Values.platformBackend.service.port }}
 - name: FF_TRACE_TIERS_ENABLED
-  value: {{ .Values.config.ttl.enabled }}
+  value: {{ .Values.config.ttl.enabled | quote }}
 - name: FF_UPGRADE_TRACE_TIER_ENABLED
-  value: {{ .Values.config.ttl.upgrade_enabled }}
+  value: {{ .Values.config.ttl.upgrade_enabled | quote}}
```
langchain-infra, line 206: nit space

PR #118, opened 2024-07-04
fix: add commonEnv and allow turning off autorestart

langchain-infra:


charts/langsmith/templates/queue/deployment.yaml
```diff
@@ -23,7 +23,9 @@ spec:
   template:
     metadata:
       annotations:
+      {{- if .Values.backend.deployment.autoRestart }}
```
bvs-langchain, line 26: Should this be queue instead of backend?

charts/langsmith/values.yaml
```diff
@@ -8,6 +8,8 @@ fullnameOverride: ""
 commonAnnotations: {}
 # -- Labels that will be applied to all resources created by the chart
 commonLabels: {}
+# -- Common environment variables that will be applied to all deployments/statefulsets created by the chart. Be careful not to override values already specified by the chart.
```
bvs-langchain, line 11: What happens if overridden? Is there concrete precedence and this is higher?

bvs-langchain (approved):

PR #123, opened 2024-07-10
Add stats queries to mirror what we have in datadog for CH health

barberscott:
Add queries for overall clickhouse health and per-query stats.  

Docs pending so draft for now.


charts/langsmith/scripts/generate_clickhouse_stats.sh
```diff
@@ -0,0 +1,165 @@
...
+            cluster_flag_passed=true
+            shift
+            ;;
+        *)
+            if [ -z "$clickhouse_url" ]; then
+                clickhouse_url="$1"
+                shift
+            else
+                echo "Unknown argument: $1"
+                echo "Usage: $0 <clickhouse_url> [--debug] [--ssl] [--days <number_of_days>] [--interval <number_of_hours>] [--output <path/to/file.csv>] [--cluster]"
```
bvs-langchain, line 56: maybe worth moving this and the 2 lines below to a `print_usage_and_exit()` fn for reuse

charts/langsmith/scripts/generate_clickhouse_stats.sh
```diff
@@ -0,0 +1,165 @@
+#!/bin/sh
+
+# Function to generate a fake UUID in lowercase
+generate_uuid() {
```
bvs-langchain, line 4: is this used?

bvs-langchain (approved):

charts/langsmith/scripts/generate_clickhouse_stats.sh
barberscott: Ah copy/pasta from the bones.  It is not.  Removing...

PR #125, opened 2024-07-12
fix: add blobstorage config

langchain-infra:
Add the ability to configure s3 using access key/secret. Workload identity support coming soon

### DIFF
--- a/charts/langsmith/templates/_helpers.tpl
+++ b/charts/langsmith/templates/_helpers.tpl
@@ -215,29 +215,29 @@ Template containing common environment variables that are used by several servic
 - name: FF_ORG_CREATION_DISABLED
   value: {{ .Values.config.orgCreationDisabled | quote }}
 {{- end }}
-{{- if .Values.config.s3Storage.enabled }}
+{{- if .Values.config.blobStorage.enabled }}
 - name: FF_S3_STORAGE_ENABLED
-  value: {{ .Values.config.s3Storage.enabled | quote }}
+  value: {{ .Values.config.blobStorage.enabled | quote }}
 - name: S3_BUCKET_NAME
-  value: {{ .Values.config.s3Storage.bucketName | quote }}
+  value: {{ .Values.config.blobStorage.bucketName | quote }}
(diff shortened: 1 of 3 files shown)


bvs-langchain (approved):

charts/langsmith/templates/_helpers.tpl
```diff
@@ -171,8 +171,6 @@ Template containing common environment variables that are used by several servic
       key: clickhouse_tls
 - name: LOG_LEVEL
   value: {{ .Values.config.logLevel }}
-- name: FF_WORKSPACE_SCOPE_ORG_INVITES_ENABLED
```
bvs-langchain, line 174: nit unrelated to your PR, maybe all non-conditional env vars should just be at the top of `commonEnv`?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -210,10 +209,36 @@ Template containing common environment variables that are used by several servic
 - name: TRACE_TIER_TTL_DURATION_SEC_MAP
   value: "{ \"longlived\": {{ .Values.config.ttl.ttl_period_seconds.longlived }}, \"shortlived\": {{ .Values.config.ttl.ttl_period_seconds.shortlived }} }"
 {{- end }}
+- name: FF_WORKSPACE_SCOPE_ORG_INVITES_ENABLED
+  value: {{ .Values.config.workspaceScopeOrgInvitesEnabled | quote }}
 {{- if .Values.config.orgCreationDisabled }}
 - name: FF_ORG_CREATION_DISABLED
   value: {{ .Values.config.orgCreationDisabled | quote }}
 {{- end }}
+{{- if .Values.config.blobStorage.enabled }}
```
bvs-langchain, line 218: is it assumed that this is AWS and not another cloud provider for now? or the `S3` env vars will get used for all providers?

bvs-langchain (approved):

charts/langsmith/templates/_helpers.tpl
langchain-infra: fixed, i inteneded to make this one conditional as well which is why imoved it 🤦

PR #126, opened 2024-07-12
Add SH script to delete an entire organization including all traces and workspaces

barberscott:
Add script and docs for removing an entire organization from a self-hosted instance.


langchain-infra (approved):
nice

barberscott:
Making it so...

PR #129, opened 2024-07-16
fix: bump chart to 0.6.32

langchain-infra:


akira (approved):

charts/langsmith/templates/_helpers.tpl
```diff
@@ -203,6 +203,11 @@ Template containing common environment variables that are used by several servic
   value: http://{{- include "langsmith.fullname" . }}-{{.Values.platformBackend.name}}:{{ .Values.platformBackend.service.port }}
 - name: SMITH_BACKEND_ENDPOINT
   value: http://{{- include "langsmith.fullname" . }}-{{.Values.backend.name}}:{{ .Values.backend.service.port }}
+- name: X_SERVICE_AUTH_JWT_SECRET
```
bvs-langchain, line 206: this doesn't match `api_key_salt`

bvs-langchain: oh I see the default is to just reuse the salt

langchain-infra: I'm just reusing api key salt here because making this a separate param would be a breaking change. Will add that to 0.7

PR #141, opened 2024-08-09
Add comments to queries describing their use, tweak docs

barberscott:
Adding a bunch of queries to calculate requested metrics to the support queries folder.

### DIFF
--- a/charts/langsmith/scripts/support_queries/clickhouse/ch_get_current_trace_counts_by_ws_monthly.sql
+++ b/charts/langsmith/scripts/support_queries/clickhouse/ch_get_current_trace_counts_by_ws_monthly.sql
@@ -1,0 +1,10 @@
+-- This query retrieves the count of monthly traces by workspace ID
+-- This count excludes deleted traces
+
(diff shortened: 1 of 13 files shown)


charts/langsmith/scripts/support_queries/clickhouse/ch_get_current_trace_counts_by_ws_monthly.sql
```diff
@@ -0,0 +1,10 @@
+-- This query retrieves the count of monthly traces by workspace ID
+-- This count excludes deleted traces
+
+select toStartOfInterval(inserted_at, interval 1 month) as ts, 
+    tenant_id as workspace_id, 
+    count(distinct id) as trace_count
+from default.runs
```
bvs-langchain, line 7: I don't see an option for setting the ClickHouse schema in helm, so this shouldn't be an issue, but just a note that if we do enable that in the future this schema will need to be an input

bvs-langchain (approved):

PR #143, opened 2024-08-12
Add script for checking status of the pvc

barberscott:
Adding support query to the Helm repo for checking the amount of space and free space on the Clickhouse volume.


langchain-infra (approved):

langchain-infra (approved):
is checking df -ah not just better here?

barberscott:
I suppose if you have kubectl access to the pod directly (since you probably do in order to do the redirect) you could run that.  This just works in the context of the other various scripts we have up there.

PR #159, opened 2024-09-19
chore: bump versions and add topology spread constraints

### LINKED ISSUE #158: add topologySpreadConstraints to deployments
I would like to request that you add topologySpreadConstraints to the deployments so that I can specify minDomains. For reliability, we typically set a minimum of two AZs in our EKS clusters. This is a more efficient method than using pod anti-affinity and has been the recommended approach since k8s 1.25.

langchain-infra:
Closes #158

### DIFF
--- a/charts/langsmith/examples/read_only_config.yaml
+++ b/charts/langsmith/examples/read_only_config.yaml
@@ -163,9 +163,9 @@ clickhouse:
   statefulSet:
     podSecurityContext:
       runAsNonRoot: true
-      runAsUser: 1000
-      runAsGroup: 1000
-      fsGroup: 1000
+      runAsUser: 101
+      runAsGroup: 101
+      fsGroup: 101
     securityContext:
       capabilities:
         drop:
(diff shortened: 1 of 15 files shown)


charts/langsmith/examples/read_only_config.yaml
```diff
@@ -163,9 +163,9 @@ clickhouse:
   statefulSet:
     podSecurityContext:
       runAsNonRoot: true
-      runAsUser: 1000
-      runAsGroup: 1000
-      fsGroup: 1000
+      runAsUser: 101
```
bvs-langchain, line 166: why this change?

bvs-langchain (approved):
lgtm, rerun helm-docs as well?

charts/langsmith/examples/read_only_config.yaml
langchain-infra: was incorrect, we use the right value in ci but not in the example

PR #162, opened 2024-09-30
feat: enable oauth in mixed auth mode

bvs-langchain:


charts/langsmith/values.yaml
```diff
@@ -82,9 +83,15 @@ config:
   personalOrgsDisabled: false
   # -- Enable Workspace Admins to invite users to the org and workspace.
   workspaceScopeOrgInvitesEnabled: false
+
+  # -- Base URL of the LangSmith installation. Used for redirects.
+  baseUrl: ""
```
langchain-infra: can we call this hostname? believe this is the param we call it in the ingress

langchain-infra (approved):
lgtm 1 cmt

PR #164, opened 2024-10-23
feat: prevent duplicate env var keys

bvs-langchain:
duplicate env vars cause strange behavior with Helm: https://github.com/fluxcd/helm-controller/issues/539. this can result in removing environment variables accidentally.

this change adds checks to fail fast if duplicate environment variables are detected in the `langsmith` chart


langchain-infra (approved):
lgtm

PR #175, opened 2024-11-08
 add /v1/platform

lc-arjun:


charts/langsmith/templates/frontend/config-map.yaml
```diff
@@ -39,6 +39,15 @@ data:
                 return 200;
          }
 
+        location ~ ^/v1/platform/ {
```
langchain-infra, line 42: i will do this when i bump the next release so you can ignore for now(will need the new images after release)

lc-arjun: so no need to merge this right now?

langchain-infra: nope, we peg to a specific version

PR #176, opened 2024-11-08
feat: enable Azure blob storage

bvs-langchain:


charts/langsmith/templates/_helpers.tpl
```diff
@@ -272,6 +272,13 @@ Template containing common environment variables that are used by several servic
 {{- if .Values.config.blobStorage.enabled }}
 - name: FF_S3_STORAGE_ENABLED
   value: {{ .Values.config.blobStorage.enabled | quote }}
+- name: FF_BLOB_STORAGE_ENABLED
```
langchain-infra, line 275: Will the old values still work?

charts/langsmith/values.yaml
```diff
@@ -107,18 +107,28 @@ config:
...
     accessKey: ""
     accessKeySecret: ""
     bucketName: ""
     apiURL: "https://s3.us-west-2.amazonaws.com"
+    # The following blob storage configuration values are for Azure and require blobStorage.engine = "Azure"
+    azureStorageAccountName: ""
+    azureStorageAccountKey: ""
+    azureStorageContainerName: ""
+    azureStorageConnectionString: ""
+    azureStorageSASToken: ""
```
langchain-infra, line 130: Do we need to support all these auth modes? E.g wouldn't really ever expect someone to use SAS or connection string (that's like presignrd url right?)

langchain-infra (approved):

charts/langsmith/templates/_helpers.tpl
bvs-langchain: old values are used for now, yeah. will start using the new ones after this is released

langchain-infra: sorry i mean like for people not using helm, will the old variables still be read

charts/langsmith/values.yaml
bvs-langchain: I think connection string is useful for the flexibility - it isn't specifically presigned URL, could use account name and key in the string. would prefer to keep that one.
probably fine to remove SAS token and add support only if needed

bvs-langchain: removed

langchain-infra: if we do wanna support connection string via secret could we just pass these in via connectionString always? Seems like that parameter should also allow you to specify accountKey or SAStoken? Normal   NotTriggerScaleUp  55s   cluster-autoscaler  pod didn't trigger scale-up: 3 node(s) didn't match Pod's node affinity/selector

langchain-infra: oops sorry meant accountKey rather than connectionString, think connectionString we should keep for sure since seems like its kind of like apiUrl?

bvs-langchain: kind of nice to construct the URL for them with accountKey and secret imo, even if that's just for teams testing out on Azure

charts/langsmith/templates/_helpers.tpl
bvs-langchain: ah right, probably worth respecting the old values until 0.9?

langchain-infra: yea though i guess that can just be a backend thing to read both

charts/langsmith/values.yaml
langchain-infra: ah gotcha, what happens if both are specified today? i think fine maybe just add a note that x overrides this if manually specified. i just generally have a bias towards connectionString when possible since it allows you to add url params and stuff (e.g when specifying db uris)

bvs-langchain: account & key has higher precedence. added a comment here & updated the docs to call this out

PR #180, opened 2024-11-13
feat(dev): expose clickhouse cluster env var

angus-langchain:
Expose clickhouse cluster env variable for self hosted replicated clickhouse setups


charts/langsmith/templates/_helpers.tpl
```diff
@@ -200,6 +200,8 @@ Template containing common environment variables that are used by several servic
     secretKeyRef:
       name: {{ include "langsmith.clickhouseSecretsName" . }}
       key: clickhouse_tls
+- name: CLICKHOUSE_CLUSTER
+  value: {{ .Values.config.clickhouseCluster }}
```
langchain-infra, line 204: can we do .Values.clickhouse.external.clickhouseCluster(where all the other params live?)

Also need to add to the .Values file

langchain-infra (approved):

PR #182, opened 2024-11-14
fix: override options for hybrid deployments

bvs-langchain:
* add new `clickhouse.external.hybrid` option: if `true`, send all payloads to blob storage and disable search

testing: ran a `helm upgrade ... --dry-run` with both `hybrid` values and observed expected env vars


langchain-infra (commented):
one cmt

charts/langsmith/examples/blob_storage_config.yaml
```diff
@@ -3,14 +3,25 @@ config:
   apiKeySalt: "YOUR_API_KEY_SALT"
   blobStorage:
     enabled: true
+    # Set engine: "Azure" if using Azure blob storage
+    engine: "S3"
+    # Set hosting: "hybrid" if using managed ClickHouse
+    hosting: "self-hosted"
```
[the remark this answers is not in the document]
bvs-langchain: that makes sense to me, updated

charts/langsmith/values.yaml
```diff
@@ -388,6 +390,8 @@ clickhouse:
     # If enabled, use the following values to connect to an external database. This will also disable the
     # creation of a clickhouse stateful-set and service.
     enabled: false
+    # -- Either "self-hosted" or "hybrid"
+    hosting: "self-hosted"
```
langchain-infra, line 394: why not just boolean here? hybrid: true or false?

langchain-infra (approved):
nit but why not just use a boolean?

charts/langsmith/values.yaml
bvs-langchain: had been thinking we would set to `cloud` and potentially add as env var, but yeah I don't think that's necessary. will update

PR #198, opened 2024-12-10
fix: add ace-backend readonly

langchain-infra:


charts/langsmith/examples/read_only_config.yaml
```diff
@@ -3,12 +3,30 @@ config:
...
+    volumes:
+      - name: tmp
+        emptyDir: {}
+    volumeMounts:
+      - name: tmp
+        mountPath: /app/node_modules
+
 backend:
   deployment:
-    resources:
```
langchain-infra, line 8: these only need to be overridden in ci

PR #217, opened 2025-02-14
Add run routes to self hosted

angus-langchain:


charts/langsmith/templates/frontend/config-map.yaml
```diff
@@ -215,6 +215,33 @@ data:
             proxy_pass http://{{ include "langsmith.fullname" . }}-{{ .Values.platformBackend.name }}:{{ .Values.platformBackend.service.port }};
          }
 
+        location = /api/v1/runs/multipart {
```
langchain-infra, line 218: can you also add to the subdomain routing above?

langchain-infra (approved):

PR #225, opened 2025-02-23
Refactoring Helm templates

namejsjeongkr:
- I've been interested in this project lately. After reviewing Helm Chart, I found some missing codes. I suggest scalable code while being consistent. (Based on the review results, I'll modify the remaining helm codes as well.)


namejsjeongkr:
@langchain-infra @angus-langchain @suraj-langchain I'd like you to review this one !

charts/langsmith/templates/ace-backend/deployment.yaml
```diff
@@ -30,19 +32,19 @@ spec:
       app.kubernetes.io/component: {{ include "langsmith.fullname" . }}-{{ .Values.aceBackend.name }}
   template:
     metadata:
-      annotations:
       {{- if .Values.aceBackend.deployment.autoRestart }}
+      annotations:
```
langchain-infra, line 36: this would break if you try to specify annotations and dont include autoRestart right?

charts/langsmith/templates/ace-backend/service.yaml
```diff
@@ -1,29 +1,19 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "langsmith.fullname" . }}-{{ .Values.aceBackend.name }}
+  name: {{ include "langsmith.fullname" . }}-{{ .Values.backend.name }}
```
langchain-infra, line 4: should be aceBackend?

charts/langsmith/templates/ace-backend/service.yaml
```diff
@@ -1,29 +1,19 @@
...
-  loadBalancerSourceRanges:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
-  loadBalancerIP: {{ .Values.aceBackend.service.loadBalancerIP }}
-  ports:
-    - name: ace
-      port: {{ .Values.aceBackend.service.port }}
-      targetPort: ace
-      protocol: TCP
+  {{- toYaml .Values.aceBackend.service.spec | nindent 2 }}
```
langchain-infra, line 16: i think I prefer specifying the service spec directly? what is the benefit of doing it this way?

langchain-infra (commented):
formatting changes and hpa changes look good to me! one question on the service spec

charts/langsmith/templates/ace-backend/deployment.yaml
namejsjeongkr: Yes, it is. If you specify annotations in the ace-backend/deployment.yaml file and do not include autoRestart, you may experience problems.

In the template, the annotations section is designed to be conditionally rendered.

Only when the autoRestart value is true is 'rollme' annotation supposed to be added.

If the user does not include autoRestart while specifying annotations, 'rollme' annotation may be missing.

This can cause the auto-start function to fail to function as intended.

charts/langsmith/templates/ace-backend/service.yaml
namejsjeongkr: correct :)

namejsjeongkr:
@langchain-infra 👀

charts/langsmith/templates/ace-backend/deployment.yaml
langchain-infra: no but we need the annotation top levle block to be added regardless right? instead of including it in the if?

charts/langsmith/templates/ace-backend/service.yaml
langchain-infra: i think the issue is that this would be a breakin change to people who are already overriding this. Would prefer not to make this change. The rest look fine to me!

charts/langsmith/templates/ace-backend/deployment.yaml
namejsjeongkr: You're absolutely right. The current implementation has a potential issue. The way it's structured, if .Values.aceBackend.deployment.autoRestart is false, but the user has specified other annotations, those annotations might not be applied correctly. The annotation block should be added regardless, with only the specific rollme annotation being conditional.

charts/langsmith/templates/ace-backend/service.yaml
namejsjeongkr: Then, What if I'd like to deploy this charts on EKS, How to write this one ? You know, AWS EKS Can't assign spec.loadBalancerSourceRanges, loadBalancerIP and so on.

langchain-infra: my understanding is they will just be empty?

namejsjeongkr: Please check it out https://github.com/langchain-ai/helm/pull/225/commits/6fcf9728acc3bb5096779c960fa8f3330c565c22

namejsjeongkr: I tried to keep most of spec attributes and delete only for specific loadBalancer !

namejsjeongkr: @langchain-infra I'm looking forward to your reply !

namejsjeongkr:
@langchain-infra @bvs-langchain @angus-langchain @suraj-langchain Please review this one. And if you don't want me to contribute, please let me know.

charts/langsmith/templates/backend/postgres-migrations.yaml
```diff
@@ -49,9 +49,7 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
           env:
-            {{- with $envVars }}
-              {{- toYaml . | nindent 12 }}
-            {{- end }}
+            {{- toYaml $envVars | nindent 12 }}
```
langchain-infra, line 52: Is there a reason to remove the with?

charts/langsmith/templates/clickhouse/service.yaml
```diff
@@ -5,30 +5,16 @@ metadata:
...
-  ports:
-    - name: ch
-      port: {{ .Values.clickhouse.service.httpPort }}
-      targetPort: ch
-      protocol: TCP
-    - name: ch-native
-      port: {{ .Values.clickhouse.service.nativePort }}
-      targetPort: ch-native
-      protocol: TCP
+  {{- toYaml .Values.clickhouse.service.spec | nindent 2 }}
```
langchain-infra, line 17: As I mentioned can we avoid this change? As it will break people using the existing config

langchain-infra (commented):
Sorry for the delay! Just one comment on the service change as it would be a breaking change for people.

charts/langsmith/templates/backend/postgres-migrations.yaml
namejsjeongkr: @langchain-infra There are several good reasons to remove the {{- with $envVars }} construct and use {{- toYaml $envVars | nindent 12 }} directly:

Simplified Logic:
The with statement skips the entire block if the variable is empty or undefined. For environment variables, which are typically critical configurations, using toYaml directly ensures they're always processed explicitly, even if empty.

Reduced Error Potential:
When using the with construct, if $envVars happens to be empty, the entire environment variable block would be omitted from the rendered template. This could lead to containers failing to start properly with difficult-to-debug errors.

Improved Readability:
Nested with statements can decrease code readability. Direct use of toYaml makes the template more straightforward and easier to understand.

Consistency:
Environment variables are essential for container operation. Using toYaml directly guarantees that the env section is always included in the template, maintaining a consistent structure.

Easier Debugging:
When issues occur, the with construct can make debugging more difficult since it conditionally includes blocks. Direct use of toYaml makes it easier to trace the state of variables during template rendering.

These advantages are particularly important for critical configurations like environment variables that directly impact application behavior. The direct approach provides more predictable template rendering and reduces potential sources of errors.

charts/langsmith/templates/clickhouse/service.yaml
namejsjeongkr: @langchain-infra I've updated it in the same form as the aceBackend code.

charts/langsmith/templates/backend/postgres-migrations.yaml
langchain-infra: but wont this break when the env section is empty?

namejsjeongkr: No, it won't break when the env section is empty. Using {{- toYaml $envVars | nindent 12 }} directly is actually safer than the previous with construct.

When $envVars is empty:

With the direct approach (toYaml $envVars), it will simply render an empty string, but the env: key will still be present in the YAML. This maintains the structure of the document.

With the previous approach using {{- with $envVars }}, if $envVars was empty, the entire env: section would be omitted, which could potentially cause issues if other parts of the application expect this section to exist.

The direct approach ensures that the env: key is always present in the output YAML, maintaining a consistent structure regardless of whether there are environment variables defined or not. This is generally considered a more robust pattern in Helm templates.

namejsjeongkr: @langchain-infra

charts/langsmith/values.yaml
```diff
@@ -583,11 +576,17 @@ frontend:
     labels: {}
     annotations: {}
   service:
-    type: LoadBalancer
-    httpPort: 80
-    httpsPort: 443
-    labels: {}
-    annotations: {}
+    spec:
```
langchain-infra, line 579: these changes should be removed right?

charts/langsmith/values.yaml
```diff
@@ -441,12 +436,6 @@ clickhouse:
     securityContext: {}
     resources:
 # For a production environment, we recommend increasing the CPU and memory limits to something like 16 cores and 64Gi.
-#      limits:
```
langchain-infra, line 444: can we keep this?

charts/langsmith/values.yaml
namejsjeongkr: That's right !

namejsjeongkr: @langchain-infra That's right !

charts/langsmith/values.yaml
namejsjeongkr: @langchain-infra Sure !

namejsjeongkr:
@langchain-infra Can you execute workflow again ?

charts/langsmith/values.yaml
```diff
@@ -226,15 +224,12 @@ aceBackend:
     port: 1987
     labels: {}
     annotations: {}
-    loadBalancerSourceRanges: []
```
langchain-infra, line 229: can we avoid removing these for the service?

charts/langsmith/values.yaml
```diff
@@ -209,8 +209,6 @@ aceBackend:
     terminationGracePeriodSeconds: 30
   autoscaling:
     enabled: false
-    # If enabled, use the following values to configure the HPA. You can also use your own HPA configuration by not creating an HPA.
```
langchain-infra, line 212: why remove this comment?

charts/langsmith/values.yaml
```diff
@@ -505,12 +498,15 @@ clickhouse:
     annotations: {}
   service:
     type: ClusterIP
-    httpPort: 8123
```
langchain-infra, line 508: i think this change needs to be reverted since we arent changing the service spec anymore?

charts/langsmith/values.yaml
```diff
@@ -590,10 +586,15 @@ frontend:
     annotations: {}
   service:
     type: LoadBalancer
-    httpPort: 80
-    httpsPort: 443
-    labels: {}
-    annotations: {}
+    ports:
```
langchain-infra, line 589: same here

namejsjeongkr:
@langchain-infra Plz review

langchain-infra:
Closing since not active

PR #243, opened 2025-03-12
feat: add langgraph operator chart

langchain-infra:
Closes INF-206


charts/langgraph-operator/templates/crds.yaml
```diff
@@ -0,0 +1,345 @@
+{{- if .Values.config.createCRDs }}
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.1
+  name: lgps.apps.langchain.ai
```
andrewnguonly, line 7: What's the impact if we rename LangGraph Platform?

charts/langgraph-operator/values.yaml
```diff
@@ -0,0 +1,58 @@
+# Default values for the langsmith helm chart. Refer to documentation on individual values for help with configuration.
```
andrewnguonly, line 1: Do some of these comments need to be updated? Update `langsmith` reference?

andrewnguonly (approved):

charts/langgraph-operator/values.yaml
langchain-infra: yes

PR #260, opened 2025-03-25
fix(Langsmith): support deploying operator with chart

langchain-infra:


charts/langsmith/templates/host-backend/deployment.yaml
```diff
@@ -1,6 +1,8 @@
 {{- define "hostBackendEnvVars" -}}
 - name: PORT
   value: {{ .Values.hostBackend.containerPort | quote }}
+- name: "HOST_COMPUTE_PLATFORM"
+  value: "k8s_operator"
```
andrewnguonly, line 5: We should also update the default value of `HOST_COMPUTE_PLATFORM` in the Python code. Then, setting this environment variable won't be necessary.

charts/langsmith/values.yaml
```diff
@@ -1,4 +1,4 @@
-# Default values for the langsmith helm chart. Refer to documentation on individual values for help with configuration.
+# Default values for the langsmith ukm chart. Refer to documentation on individual values for help with configuration.
```
andrewnguonly, line 1: Typo?

andrewnguonly (approved):

charts/langsmith/templates/host-backend/deployment.yaml
langchain-infra: fixed

charts/langsmith/values.yaml
langchain-infra: fixed

PR #275, opened 2025-04-09
feat: Experimental Search support

xornivore:
## Motivation

* Provides settings and configuration for experimental search features in self-hosted deployments.

## Changes

* Can be enabled with `config.fullTextSearch.indexing.enabled`
* Reuses postgres DB already in the chart for metastore 
* Requires minimal config for object storage, example:

```
..
  config:
    default_index_root_uri: s3://<bucket>/indexes
    storage:
      s3:
        endpoint: "https://s3.us-west-2.amazonaws.com"
        region: us-west-2
...
  indexer:
    statefulSet:
      extraEnvFrom:
      - secretRef:
          name: <bucket-secret>

  searcher:
    statefulSet:
      extraEnvFrom:
      - secretRef:
          name: <bucket-secret>


  metastore:
    postgres:
      shareWithLangSmith: false
      existingSecretName: <metastore-secret>
      connectionUrlSecretKey: metastore_connection_url

```

Where:
* `<bucket>` - S3 bucket for storing indexes
* `<bucket-secret>` - S3 credentials - key id/secret env vars
* `<metastore-secret>` - secret containing metastore connection URL


charts/langsmith/values.yaml
```diff
@@ -1408,3 +1411,352 @@ redis:
...
+  image:
+    repository: quickwit/quickwit
+    pullPolicy: IfNotPresent
+    tag: edge
+
+  imagePullSecrets: []
+  nameOverride: ""
+  fullnameOverride: ""
+
+  # Set the Kuberentes cluster domain if not default. It's used to build URLs for the services.
```
langchain-infra, line 1427: dont think we need this / would do it in the top level chart

charts/langsmith/templates/quickwit/service.yaml
```diff
@@ -0,0 +1,193 @@
+{{- if .Values.config.fullTextSearch.enabled -}}
```
langchain-infra, line 1: nit, can we separate this out as 3 separate services? e.g janitor-service, searcher-service, etc?

langchain-infra (commented):
a bunch of nits related to following the configs from our previous charts. I'd like to keep our specs in the values.yaml kind of the same for each service. Would just put the following sections under quickwit:

```
quickwit:
  controlPlane:
  indexer:
  janitor:
  metastore:
  searcher:
  updateRunsIndexJob:
```
  
  and eahc of these follows the configs of an individual service like frontend. Does that make sense? I do see a lot of the logic is duplicated so okay with some of that being consolidated but people always want to add specific tolerations, labels etc to individual components so i like being granular when possible.

charts/langsmith/templates/_helpers.tpl
```diff
@@ -458,3 +468,179 @@ Template containing common environment variables that are used by several servic
...
+  httpGet:
+    path: /health/livez
+    port: rest
+readinessProbe:
+  httpGet:
+    path: /health/readyz
+    port: rest
+{{- end }}
+
+# {{- include "quickwit.liveness-probe" . }}
```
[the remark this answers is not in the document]
xornivore: nope, will clean up, thank you.

charts/langsmith/templates/_helpers.tpl
```diff
@@ -458,3 +468,179 @@ Template containing common environment variables that are used by several servic
...
+  value: "$(POD_IP)"
+- name: QW_CLUSTER_ENDPOINT
+  value: http://{{ include "quickwit.fullname" $ }}-metastore.{{ $.Release.Namespace }}.svc.{{ .Values.quickwit.clusterDomain }}:7280
+{{- range $key, $value := .Values.quickwit.environment }}
+- name: "{{ $key }}"
+  value: "{{ $value }}"
+{{- end }}
+{{- end }}
+
+{{- define "quickwit.probes" -}}
```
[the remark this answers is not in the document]
xornivore: Sounds good, will move them there, kind of went in the opposite direction, but would be good to just have a single spot since they are the same across components.

charts/langsmith/templates/quickwit/control-plane-deployment.yaml
```diff
@@ -0,0 +1,104 @@
+{{- if .Values.config.fullTextSearch.enabled -}}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "quickwit.fullname" . }}-control-plane
```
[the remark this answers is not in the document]
xornivore: Yeah, good idea.

charts/langsmith/templates/_helpers.tpl
```diff
@@ -458,3 +468,179 @@ Template containing common environment variables that are used by several servic
...
+Quickwit name
+*/}}
+{{- define "quickwit.name" -}}
+{{ include "langsmith.name" . }}-{{ .Values.quickwit.name }}
+{{- end }}
+
+{{/*
+Fully qualified Quickwit name
+*/}}
+{{- define "quickwit.fullname" -}}
```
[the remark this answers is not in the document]
xornivore: can clean up for sure

charts/langsmith/templates/_helpers.tpl
```diff
@@ -458,3 +468,179 @@ Template containing common environment variables that are used by several servic
...
+{{- define "quickwit.serviceAccountName" -}}
+{{- if .Values.quickwit.serviceAccount.create }}
+{{- default (include "quickwit.fullname" .) .Values.quickwit.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.quickwit.serviceAccount.name }}
+{{- end }}
+{{- end }}
+
+{{/*
+Quickwit ports
```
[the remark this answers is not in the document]
xornivore: that sounds good, it's really just the two ports but can see it being useful if there is a conflict

charts/langsmith/templates/quickwit/service.yaml
xornivore: I'd like it to be as simple as possible, but curious what you are thinking could be the use case to separate, e.g. they do support searcher without indexer but it serves a purpose of readonly searching on prebuilt bucket with index. Are you thinking to disable janitor (no deletes) to minimize the footprint? We could probably split off:
`fullTextSearch.indexing.enabled` - runs `searcher` + `indexer`
`fullTextSearch.deletes.enabled` - runs `janitor` - we currently don't support it anyway so can disable for now.

charts/langsmith/values.yaml
```diff
@@ -1408,3 +1411,352 @@ redis:
     name: ""
     labels: {}
     annotations: {}
+
+quickwit:
+  name: quickwit
+
+  image:
```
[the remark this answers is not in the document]
xornivore: Can look into that some more. Reusing `imagePullSecrets` makes sense.

charts/langsmith/values.yaml
xornivore: `clusterDomain` is used to form service endpoints.

charts/langsmith/templates/_helpers.tpl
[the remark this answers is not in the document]
xornivore: It actually does help to have this in a couple of places, e.g. in headless service definition and just prefixing all things quickwit, but good idea to have overridable names for components.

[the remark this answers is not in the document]
xornivore: Refactored a bit differently - defined `quickwit.configMapName` and `quickwit.clusterName`

charts/langsmith/templates/quickwit/service.yaml
xornivore: This is updated.

charts/langsmith/values.yaml
```diff
@@ -54,7 +54,10 @@ images:
     repository: "docker.io/clickhouse/clickhouse-server"
     pullPolicy: Always
     tag: "24.8"
-
+  quickwitImage:
+    repository: quickwit/quickwit
+    pullPolicy: IfNotPresent
+    tag: edge
```
bvs-langchain, line 60: can we use a specific version here or needs to be `edge`?

bvs-langchain (approved):

charts/langsmith/Chart.yaml
```diff
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
-version: 0.9.37
+version: 0.9.38
```
langchain-infra, line 8: i think this might need to be 0.9.40

charts/langsmith/templates/quickwit/control-plane-deployment.yaml
```diff
@@ -0,0 +1,106 @@
...
+    {{- with .Values.quickwit.controlPlane.deployment.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- include "quickwit.annotations" . | nindent 4 }}
+    {{- with .Values.quickwit.controlPlane.deployment.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  replicas: 1
```
langchain-infra, line 18: should we allow setting this? or is it a singleton?

langchain-infra (approved):

charts/langsmith/templates/quickwit/job-update-runs-index.yaml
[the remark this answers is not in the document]
xornivore: yes, makes sense, more consistent

charts/langsmith/Chart.yaml
xornivore: sounds good, will update

charts/langsmith/templates/quickwit/control-plane-deployment.yaml
xornivore: I haven't had a reason to set it to more than a single replica, but will confirm if it actually supports it. For now I'd say we need to remove it and allow configuring once there is a reason.

PR #279, opened 2025-04-14
feat: use shared ingress for self-hosted lgp control plane and dataplane

langchain-infra:


charts/langsmith/templates/_helpers.tpl
```diff
@@ -330,10 +330,12 @@ Template containing common environment variables that are used by several servic
   value: "host"
 - name: HOST_WORKER_RECONCILIATION_CRON_ENABLED
   value: "true"
-- name: LANGCHAIN_ENDPOINT
```
bvs-langchain, line 333: is this set elsewhere or just unused?

charts/langsmith/templates/platform-ingress.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if .Values.config.langgraphPlatform.enabled }}
```
bvs-langchain, line 1: nit: maybe rename file to `langgraph-platform-ingress.yaml` or `lg-...` since we have `platform-backend`

charts/langsmith/values.yaml
```diff
@@ -1025,14 +1032,14 @@ operator:
                 protocol: TCP
               livenessProbe:
                 httpGet:
-                  path: /ok?check_db=1
+                  path: /${name}/ok?check_db=1
```
bvs-langchain, line 1035: OOC where's the corresponding change that prefixes this path?

charts/langsmith/templates/_helpers.tpl
langchain-infra: we want to use the default

charts/langsmith/values.yaml
langchain-infra: the other pr i sent you this morning that sets mount_prefix

langchain-infra: https://github.com/langchain-ai/langchainplus/pull/9167

charts/langsmith/templates/platform-ingress.yaml
langchain-infra: good call

charts/langgraph-dataplane/templates/ingress.yaml
```diff
@@ -0,0 +1,30 @@
...
+    {{- end }}
+spec:
+  ingressClassName: {{ .Values.ingress.ingressClassName }}
+  {{- with .Values.ingress.tls }}
+  tls:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  defaultBackend:
+    service:
+      name: some-placeholder-service  # or something that returns 404
```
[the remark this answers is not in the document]
langchain-infra: no they dont need to set anything herre but it fails validation if you dont at least specify a placeholder default

charts/langsmith/templates/platform-ingress.yaml
langchain-infra: updated

PR #291, opened 2025-04-24
Update app.kubernetes.io labels to be consistent across templates

nazzzzz:
This label is already hard-coded in the deployment templates, it's causing duplication issues when we template the chart with kustomize. LMK if you need more info

### DIFF
--- a/charts/langgraph-dataplane/templates/listener/deployment.yaml
+++ b/charts/langgraph-dataplane/templates/listener/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   name: {{ include "langgraphDataplane.fullname" . }}-{{ .Values.listener.name }}
   labels:
     {{- include "langgraphDataplane.labels" . | nindent 4 }}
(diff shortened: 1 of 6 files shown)


langchain-infra (approved):

langchain-infra (changes requested):
which deplyoment already included this? i think this will actually break existing deployments by changing th selector labels (which i believe is immutable)

charts/langgraph-dataplane/templates/listener/deployment.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   name: {{ include "langgraphDataplane.fullname" . }}-{{ .Values.listener.name }}
   labels:
     {{- include "langgraphDataplane.labels" . | nindent 4 }}
+    app.kubernetes.io/component: {{ include "langgraphDataplane.fullname" . }}-{{ .Values.operator.name }}
```
langchain-infra, line 9: should be .listener.name

nazzzzz: Whoops

langchain-infra (approved):

PR #295, opened 2025-04-29
fix: sync CSP for subdomain ingress

bvs-langchain:


langchain-infra (approved):

langchain-infra (commented):
do you need to bump chart version

PR #300, opened 2025-05-01
consistent checksum config annotation

nazzzzz:


langchain-infra:
Closing since fixed in separate pr

PR #305, opened 2025-05-06
feat: add pdb to langgraph cloud chart

langchain-infra:


charts/langgraph-cloud/values.yaml
```diff
@@ -95,6 +95,9 @@ apiServer:
     name: ""
     labels: {}
     annotations: {}
+  pdb:
```
bvs-langchain, line 98: these are missing `maxUnavailable`

bvs-langchain (approved):

PR #314, opened 2025-05-23
bump helm charts

angus-langchain:


charts/langsmith/Chart.yaml
```diff
@@ -6,4 +6,4 @@ maintainers:
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
 version: 0.10.17
```
langchain-infra, line 8: can you also bump chart version

langchain-infra (approved):

PR #320, opened 2025-05-28
Support specifying one registry for all images in helm

romain-priour-lc:
Tested by manually deploying on EKS both with and without default set


charts/langsmith/values.yaml
```diff
@@ -12,6 +12,7 @@ commonLabels: {}
 commonEnv: []
 
 images:
+  defaultRegistry: ""
```
langchain-infra, line 15: can we add a comment specifying that if this is used, we will concatenate it to all images?

charts/langsmith/templates/operator/deployment.yaml
```diff
@@ -48,7 +48,7 @@ spec:
         args:
           - --leader-elect
           - --health-probe-bind-address=:8081
-        image: "{{ .Values.images.operatorImage.repository }}:{{ .Values.images.operatorImage.tag | default "latest" }}"
+        image: {{ include "langsmith.image" (dict "Values" .Values "Chart" .Chart "component" "operatorImage") | quote }}
```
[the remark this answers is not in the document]
romain-priour-lc: from my understanding you can pass dicts or lists to pass multiple values. Dicts seem better in my eyes because you can refer to the values by name instead of index

charts/langgraph-cloud/templates/_helpers.tpl
```diff
@@ -143,3 +143,15 @@ the user or some other secret provisioning mechanism
...
+{{/*
+Creates the image reference used for LangGraph Cloud deployments. If registry is specified, concatenate it, along with a '/'.
+*/}}
+{{- define "langGraphCloud.image" -}}
+{{- $imageConfig := index .Values.images .component -}}
+{{- if .Values.images.registry -}}
+{{ .Values.images.registry }}/{{ $imageConfig.repository }}:{{ $imageConfig.tag | default .Chart.AppVersion }}
+{{- else -}}
+{{ $imageConfig.repository }}:{{ $imageConfig.tag | default .Chart.AppVersion }}
+{{- end -}}
```
langchain-infra, line 156: nit nl

langchain-infra (approved):
lgtm!

PR #335, opened 2025-06-05
chore: add persistentVolumeClaimRetentionPolicy

langchain-infra:


romain-priour-lc (approved):
lgtm

joaquin-borggio-lc (approved):
looks good to me but can we do any quick validation that this doesnt break when no policy is provided? I think a simple helm install would quickly validate?

PR #336, opened 2025-06-06
Add HPA addtional custom metrics

namejsjeongkr:
I'd like to add custom metrics for HPA. I think it could be useful when we have to consider lots of situations.


namejsjeongkr:
@langchain-infra How can I merge this one ? Do I have to do something to pass workflow ?

PR #343, opened 2025-06-09
chore: support additional paths langgraph dataplane

langchain-infra:


charts/langgraph-dataplane/templates/ingress.yaml
```diff
@@ -27,11 +27,15 @@ spec:
   - host: {{ .Values.ingress.hostname }}
     http:
       paths:
-      - path: /
+      - path: /metrics
```
romain-priour-lc, line 30: why default to metrics now?

langchain-infra: its the only route that is actually exposed on this service, customer wants to use / to route to another domain.

romain-priour-lc (approved):
lgtm

PR #348, opened 2025-06-11
Create feedback migration k8s job

angus-langchain:
Create idle migration job for self hosted feedback migration


charts/langsmith/templates/backend/feedback-migration.yaml
```diff
@@ -0,0 +1,140 @@
+{{- if and .Values.backend.feedbackMigration .Values.backend.feedbackMigration.configMigration .Values.backend.feedbackMigration.configMigration.enabled -}}
```
langchain-infra, line 1: could we convert this to just two sections?

backend.feedbackConfigMigration
backend.feedbackMigration

Also would split this up into two files for easier grokking. With each template being created by .Values.backend.feedbackConfigMigration.enabled or .Values.backend.feedbackMigration.enabled respectively. Not needed to nest.

charts/langsmith/values.yaml
```diff
@@ -340,6 +340,41 @@ backend:
...
+      enabled: false
+      annotations: {}
+      podAnnotations: {}
+      podSecurityContext: {}
+      securityContext: {}
+      resources:
+        limits:
+          cpu: 4000m
+          memory: 8Gi
+        requests:
```
langchain-infra, line 371: this is decently heft, do we need this much?

langchain-infra (commented):
mostly lgtm! 1 nit

charts/langsmith/values.yaml
angus-langchain: we needed this much for cloud so can probably reduce slightly, but this is a large migration, we're pulling in all feedback per month into memory and processing each row individually via cpu, and then batch inserting to postgres

langchain-infra (approved):

PR #350, opened 2025-06-12
More customizations for Migration scripts and E2E trace test

romain-priour-lc:
1. Expose initContainers in all deployments and stateful sets
2. Expose many more parameters to the E2E test, including sidecar, service account
3. Allow to configure separate service accounts and override connection string for Postgres migration jobs


charts/langsmith/templates/backend/migrations-service-account.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.backend.migrations.serviceAccount.create -}}
```
langchain-infra, line 1: do we need this? I think they were just asking for a separate db user, not service account? Would prefer not to add this unless absoltuely needed? This is potentially a breaking change if users were relying on the backend SA previously

charts/langsmith/templates/backend/postgres-migrations.yaml
```diff
@@ -52,6 +56,10 @@ spec:
             {{- with $envVars }}
               {{- toYaml . | nindent 12 }}
             {{- end }}
+            {{- with .Values.backend.migrations.connectionString }}
```
langchain-infra, line 59: instead of adding it this way, could we have them add an env var and then echo it by modifying the container command if they want this? I don't think very common ask. (Main reason is that the format for this env var is actually different, we expect it to have a different extension to start)

langchain-infra (approved):
2 cmts

charts/langsmith/templates/backend/migrations-service-account.yaml
romain-priour-lc: I for confused here between Postgres and K8s service accounts. Removing this

PR #369, opened 2025-06-27
fix: use webserver health checks for faster queue startup

langchain-infra:
https://github.com/tobymao/saq/issues/44#issuecomment-1296297391 author of the lib seems to use this approach

Closes INF-584


charts/langgraph-dataplane/values.yaml
```diff
@@ -67,33 +68,30 @@ listener:
...
-          - "app.workers.queues.host_worker.settings"
-          - "--check"
+      httpGet:
+        path: /health
+        port: 8080
       failureThreshold: 6
-      periodSeconds: 60
-      timeoutSeconds: 30
+      periodSeconds: 10
+      timeoutSeconds: 1
```
joaquin-borggio-lc, line 80: Is 1 second enough time if everything scales up at once?

joaquin-borggio-lc (approved):
Just 1 comment on if the threshold is enough for large scale up

PR #389, opened 2025-07-22
Update `langgraph-dataplane` `listener` image version to the latest version

andrewnguonly:
Also, explicitly adding `LOG_LEVEL` environment variable so that it's more obvious that it can be set. Setting `LOG_LEVEL=DEBUG` is important for debugging `listener` issues/errors.


langchain-infra (approved):

langchain-infra (approved):
Need to bump chart version

PR #394, opened 2025-07-24
chore: bump default backend probe timeouts (for high load)

joaquin-borggio-lc:
Bump the default timeout for probes on the backend service from 1s to 10s.
Normally 1s is fine when the cluster is not under high load. Under high load some requets to stateful services can take longer and so the health check needs more time.

## Testing
During load testing, backend pods would crash bc of timeouts on the probes constantly. When bumping timeout to 10s, I didnt see any more restarts.


romain-priour-lc (approved):
update in README?

PR #408, opened 2025-08-05
Add support for setting `langgraphListenerId`

andrewnguonly:


charts/langgraph-dataplane/values.yaml
```diff
@@ -51,6 +51,7 @@ config:
   hostBackendUrl: "https://api.host.langchain.com"
   smithBackendUrl: "https://api.smith.langchain.com"
   langsmithWorkspaceId: ""
+  langgraphListenerId: ""
```
langchain-infra, line 54: do we need to bump the chart and make this required?

andrewnguonly: It isn't required to maintain backwards compatibility with existing hybrid deployments where the listener deploys in the current namespace.

charts/langgraph-dataplane/templates/listener/rbac.yaml
```diff
@@ -1,9 +1,13 @@
 {{- if .Values.listener.rbac.create }}
+{{- if .Values.operator.watchNamespaces }}
+{{- $watchNamespacesList := splitList "," .Values.operator.watchNamespaces }}
+{{- range $ns := $watchNamespacesList }}
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ include "listener.serviceAccountName" $ }}-role
-  namespace: {{ .Values.namespace | default .Release.Namespace }}
```
andrewnguonly, line 6: TODO: Figure out how to add these to the `watchNamespacesList`.

charts/langgraph-dataplane/templates/listener/rbac.yaml
```diff
@@ -1,17 +1,21 @@
 {{- if .Values.listener.rbac.create }}
+{{- if .Values.listener.watchNamespaces }}
```
langchain-infra, line 2: does operator need changes here as well?

charts/langgraph-dataplane/values.yaml
langchain-infra: well you do need to bump chart.yaml and probably should start setting app version= listener version

charts/langgraph-dataplane/templates/listener/rbac.yaml
andrewnguonly: I don't think so. The `operator` already supports `watchNamespaces`. But I guess it's worth documenting that `listener.watchNamespaces` and `operator.watchNamespaces` need to be equal.

charts/langgraph-dataplane/templates/_helpers.tpl
```diff
@@ -121,12 +121,16 @@ Template containing common environment variables that are used by several servic
...
+- name: HOST_WORKER_LISTENER_ID
+  value: {{ .Values.config.langgraphListenerId }}
 - name: HOSTED_K8S_ROOT_DOMAIN
   value: {{ .Values.config.rootDomain | quote }}
 - name: HOSTED_K8S_SHARED_INGRESS
   value: "true"
 - name: LOG_LEVEL
   value: "INFO"
+- name: ENABLE_LGP_DEPLOYMENT_HEALTH_CHECK
+  value: "true"
```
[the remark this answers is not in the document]
langchain-infra: well its not toggleable?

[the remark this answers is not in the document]
langchain-infra: that doesn't work because everytime you deploy it would get overwritten.

charts/langgraph-dataplane/templates/operator/deployment.yaml
```diff
@@ -100,7 +100,7 @@ spec:
         {{- end }}
         env:
           - name: WATCH_NAMESPACE
-            value: {{ .Values.operator.watchNamespaces | quote }}
+            value: {{ default .Values.operator.watchNamespaces .Values.config.watchNamespaces | quote }}
```
andrewnguonly, line 103: For backward compatibility, default to `.Values.operator.watchNamespaces` if `.Values.config.watchNamespaces` is not set.

charts/langgraph-dataplane/templates/operator/rbac.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if and .Values.operator.enabled .Values.operator.rbac.create -}}
-{{- if .Values.operator.watchNamespaces }}
-{{- $watchNamespacesList := splitList "," .Values.operator.watchNamespaces }}
+{{- if or .Values.config.watchNamespaces .Values.operator.watchNamespaces }}
+{{- $watchNamespacesList := splitList "," (default .Values.operator.watchNamespaces .Values.config.watchNamespaces) }}
```
andrewnguonly, line 3: For backward compatibility, default to .Values.operator.watchNamespaces if .Values.config.watchNamespaces is not set.

charts/langgraph-dataplane/templates/_helpers.tpl
```diff
@@ -121,12 +121,16 @@ Template containing common environment variables that are used by several servic
...
+- name: HOST_WORKER_LISTENER_ID
+  value: {{ .Values.config.langgraphListenerId }}
 - name: HOSTED_K8S_ROOT_DOMAIN
   value: {{ .Values.config.rootDomain | quote }}
 - name: HOSTED_K8S_SHARED_INGRESS
   value: "true"
 - name: LOG_LEVEL
   value: "INFO"
+- name: ENABLE_LGP_DEPLOYMENT_HEALTH_CHECK
+  value: {{ .Values.listener.enableLGPDeploymentHealthCheck | quote }}
```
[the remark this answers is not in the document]
andrewnguonly: The rule that I was following (but I don't know if this is the correct rule):
1. `config`-level is for values used across more than 1 deployment/thing.
1. Otherwise, the config can be placed in the deployment-specific config.

The inverse of this rule is applied for `watchNamespaces`. Since `watchNamespaces` is used in 2 places, it was moved to the `config`-level.

charts/langgraph-dataplane/templates/listener/rbac.yaml
```diff
@@ -1,17 +1,21 @@
 {{- if .Values.listener.rbac.create }}
+{{- if .Values.config.watchNamespaces }}
```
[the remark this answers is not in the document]
andrewnguonly: Update the logic to fall back to `.Values.namespace | default .Release.Namespace` if `.Values.config.watchNamespaces` is not set.

langchain-infra (approved):
lets run helm-docs as well?

PR #414, opened 2025-08-11
feat: bump v11

langchain-infra:
Closes INF-674
Closes INF-676
Closes INF-675


andrewnguonly (approved):

charts/langsmith/values.yaml
```diff
@@ -1396,9 +1371,7 @@ queue:
         cpu: 1000m
         memory: 2Gi
     command:
-      - "saq"
-      - "app.workers.queues.single_queue_worker.settings"
-      - "--quiet"
+      - "./queue_entrypoint.sh"
```
joaquin-borggio-lc, line 1374: I checked the entrypoint scripts real quick to see if things had changed since we created those a few weeks ago. I see here that this script has the other command when we had the lightweight health checks:
```
exec saq app.workers.queues.single_queue_worker.settings --quiet --web --port "$PORT"
```

We want to move to the lightweight ones anyway for the startup and readiness probes but just fyi that it is changing here.

langchain-infra: im down to keep the new command and move the to the new probes.

PR #415, opened 2025-08-12
feat: Route hub commits (/commits) endpoints to platform-backend

xornivore:
## Description

- Adds routing to the new Go-based implementation in `platform-backend` for `/commits` and `/api/v1/commits` endpoints improving performance and reliability of these operations.


charts/langsmith/templates/frontend/config-map.yaml
```diff
@@ -516,6 +534,23 @@ data:
             proxy_pass http://{{ include "langsmith.fullname" . }}-{{ .Values.platformBackend.name }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.platformBackend.service.port }};
         }
 
+        location = /commits {
```
langchain-infra, line 537: in sh we use /api/x (so would wanna route /api/commits and /api/v1/commits. (similar to /runs).

PR #429, opened 2025-08-27
feat: New SSO integration flow

romain-priour-lc:


charts/langsmith/templates/backend/auth-bootstrap.yaml
```diff
@@ -28,18 +29,31 @@
...
       name: {{ include "langsmith.secretsName" . }}
       key: initial_org_admin_password
+- name: INITIAL_ORG_NAME
+  value: {{ .Values.config.initialOrgName }}
+- name: DEFAULT_WORKSPACE_NAME
+  value: {{ .Values.config.defaultWorkspaceName }}
 {{- end -}}
 {{- template "validateAuthMode" . -}}
-{{- if .Values.config.basicAuth.enabled }}
+{{- if or .Values.config.basicAuth.enabled .Values.config.oauth.enabled }}
```
bvs-langchain, line 56: why this change?

charts/langsmith/values.yaml
```diff
@@ -105,15 +105,22 @@ config:
...
+
   basicAuth:
     enabled: false
+    # -- Kept for backwards compatibility. Will be removed in a future release. Please use values.config.initialOrgAdminEmail instead.
     initialOrgAdminEmail: ""
     # Either set the password and JWT secret in plaintext, or set in the secret specified by existingSecretName above.
     initialOrgAdminPassword: ""
     jwtSecret: ""
-  # -- Prevent organization creation. If using basic auth, this is set to true by default.
+  # -- Prevent organization creation. If using basic auth or oauth, this is set to true by default.
```
bvs-langchain, line 121: nit: `oauth with client secret`, same below

bvs-langchain (approved):

charts/langsmith/templates/backend/auth-bootstrap.yaml
romain-priour-lc: the {{- if .Values.config.basicAuth.enabled }} statement was terminated on the last line of the file, so auth bootstrap didn't run for oauth. 

The logic should be cleaner now

PR #431, opened 2025-08-27
chore: move to web server health checks for startup/readiness probe in queue and listener

joaquin-borggio-lc:
Similar to a past PR but only setting this quicker health chekc for startup and readiness. for liveness, we want to heavier check that is more accurate

Previous PR:
https://github.com/langchain-ai/helm/pull/369/


langchain-infra (approved):
can we also update dataplane with this one?

PR #437, opened 2025-09-09
Custom CA for self-hosted deployments

romain-priour-lc:
- Allows to mount a custom CA to all deployments, enabling TLS with custom CA for databases, etc.


charts/langsmith/values.yaml
```diff
@@ -1254,6 +1259,7 @@ postgres:
     connectionUrl: ""
     connectionUrlSecretKey: "connection_url"
     existingSecretName: ""
+    tls: false
```
langchain-infra, line 1262: nice. wondering if this will break people who already had SSL turned on?

langchain-infra (approved):

PR #440, opened 2025-09-10
Add HOST_BACKEND_ENDPOINT_PUBLIC env var if ingress provided

romain-priour-lc:
Currently, the host-backend docs show the Hostname as the internal Kubernetes endpoint, even if there is an ingress. This instead shows the DNS name of the ingress


joaquin-borggio-lc (approved):
this seems fine to me and i see this var, but maybe good to get andrews eyes as well since its very short/quick

PR #449, opened 2025-09-18
Bump helm 0.11.14 + Disable org creation and personal orgs by default

romain-priour-lc:
- Since we disable for both basic auth and oauth, this is much more intuitive.
- Users can just override the value in the chart to re-enable org creation

### DIFF
--- a/charts/langsmith/Chart.yaml
+++ b/charts/langsmith/Chart.yaml
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
-version: 0.11.15
+version: 0.11.14
 appVersion: "0.11.45"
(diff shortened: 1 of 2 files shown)


charts/langsmith/values.yaml
```diff
@@ -118,10 +118,10 @@ config:
     # Either set the password and JWT secret in plaintext, or set in the secret specified by existingSecretName above.
     initialOrgAdminPassword: ""
     jwtSecret: ""
-  # -- Prevent organization creation. If using basic auth or oauth with client secret, this is set to true by default.
-  orgCreationDisabled: false
-  # -- Disable personal orgs. Users will need to be invited to an org manually. If using basic auth or oauth with client secret, this is set to true by default.
-  personalOrgsDisabled: false
+  # -- Prevent organization creation.
+  orgCreationDisabled: true
```
bvs-langchain, line 122: the bootstrap still succeeds when this is `true`, correct?

bvs-langchain (approved):

charts/langsmith/values.yaml
bvs-langchain: I guess more accurate naming would be `userOrgCreationDisabled` - maybe worth adding to comment "Prevent organization creation by users. This includes admins."

romain-priour-lc: Tested, still succeeds. Will add a comment

PR #467, opened 2025-09-30
chore: filter K8s debugging script to LangChain resources only

nitboss:
## Summary
- Updates `get_k8s_debugging_info.sh` to filter results for only LangChain-related resources
- Allows customers to embed other resources in the same cluster without polluting debug output
- Removes `jq` dependency by using `jsonpath` for restart count check

## Changes
- **Resource filtering**: Added grep filters for `langsmith-*`, `lg-*`, and `toolbox*` prefixes to:
  - Resource summary output
  - Resource details YAML
  - Pod resource usage
  - Pod log collection
- **Simplified restart count check**: Changed from `jq` to `jsonpath` for better portability
- **Improved output messages**: Updated echo statements for clarity

## Test Plan
- [ ] Test script on namespace with mixed LangChain and non-LangChain resources
- [ ] Verify only LangChain resources appear in output files
- [ ] Test with pods that have restart counts > 0
- [ ] Confirm no `jq` dependency required

### DIFF
--- a/charts/langsmith/scripts/get_k8s_debugging_info.sh
+++ b/charts/langsmith/scripts/get_k8s_debugging_info.sh
@@ -1,17 +1,33 @@
 #!/bin/bash
 
 # We expect the namespace hosting all kubernetes resources to be passed as an argument to this script
+FILTER_REGEX=""
+FILTER_LABELS=""
+
 while [[ "$#" -gt 0 ]]; do
   case $1 in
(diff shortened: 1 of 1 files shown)


langchain-infra (commented):
can we just add an option to this chart to filter by regex or label which is moer common in k8s? people deploy Ls under different names (e.g LG platform etc).

langchain-infra (approved):

PR #470, opened 2025-10-02
feat: support overriding CSP & using nonce

bvs-langchain:
* adds `includeNonce` option to optionally replace %%NONCE%% in bundle with a nonce value to support stricter CSP
* adds `cspHeader` option using current value as default to allow overriding CSP

### DIFF
--- a/charts/langsmith/Chart.yaml
+++ b/charts/langsmith/Chart.yaml
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
-version: 0.11.21
+version: 0.11.22
 appVersion: "0.11.64"


eric-langchain (approved):

charts/langsmith/templates/frontend/config-map.yaml
```diff
@@ -411,18 +411,30 @@ data:
         proxy_send_timeout {{ .Values.frontend.proxyWriteTimeout }};
         keepalive_timeout {{ .Values.frontend.keepAliveTimeout }};
 
-        add_header Content-Security-Policy "frame-ancestors 'self'; object-src 'none'" always;
+        add_header Content-Security-Policy "{{ .Values.frontend.cspHeader }}" always;
```
eric-langchain, line 414: does user have to set: 

"script-src 'nonce-$request_id' 'strict-dynamic'; style-src 'nonce-$request_id' 'strict-dynamic';" always;

themselves?

bvs-langchain: yes, we _could_ pass through the LangSmith domain, but I think slightly preferred to have the flexibility, at least while this is under development

PR #485, opened 2025-10-22
Add commonInitContainers and toggle to disable secret creation

romain-priour-lc:


charts/langsmith/values.yaml
```diff
@@ -99,6 +101,7 @@ istioGateway:
 
 config:
   existingSecretName: ""
+  disableSecretCreation: false
```
langchain-infra: slight nit, can we make this part of the existingSecretName section for each of the dbs as well?

langchain-infra: can we add next to the existingSecretName param? and also maybe document like "if setting this, make sure you are injecting the env vars that use this secret yourself"?

PR #490, opened 2025-10-24
Add S3UsePathStyle flag

romain-priour-lc:


langchain-infra (approved):
lgtm, maybe add an example like "will call s3.us-west-2.amazonaws.com/<bucketname> vs bucketname.s3.x"?

PR #491, opened 2025-10-24
Fix: Validate required config apiKeySalt and langsmithLicenseKey are set

ben11211:


charts/langsmith/templates/validate.yaml
```diff
@@ -38,10 +38,18 @@ It only performs validation checks and fails template rendering if issues are fo
 {{- fail "Only one of ingress, gateway, or istioGateway can be enabled at the same time." -}}
 {{- end -}}
 
+{{- /* Validate langsmithLicenseKey is set */ -}}
+{{- if not .Values.config.langsmithLicenseKey -}}
+{{- fail "config.langsmithLicenseKey is required." -}}
+{{- end -}}
```
joaquin-borggio-lc: people can put the licesne key in the existing secret so we should check that the existingSecretName field is not set either

charts/langsmith/templates/validate.yaml
```diff
@@ -38,10 +38,18 @@ It only performs validation checks and fails template rendering if issues are fo
...
+
 {{- /* Validate authentication configuration */ -}}
 {{- if and .Values.config.basicAuth.enabled .Values.config.oauth.enabled -}}
 {{- fail "Basic auth and OAuth are both enabled. Please enable only one." -}}
 {{- end -}}
+{{- if and (not .Values.config.apiKeySalt ) (not .Values.config.existingSecretName) -}}
+{{- fail "config.apiKeySalt is required. This may have previously been set to your license key by default. Set this value to the license key to avoid invalidating existing API Keys. This can alternatively be set within .Values.config.existingSecretName." -}}
+{{- end -}}
```
joaquin-borggio-lc: like this yeah

PR #497, opened 2025-11-04
Add `list` and `delete` permission for pvcs to `listener`

andrewnguonly:
### Summary
New functionality in the `listener` will automatically clean up persistent volume claims (pvcs). The `listener` needs permission to `list` and `delete` these resources.

We don't need to bump the chart versions now. When they're bumped later, deployments will automatically pick up these RBAC changes with the new functionality of the `listener`.

### DIFF
--- a/charts/langsmith/templates/listener/rbac.yaml
+++ b/charts/langsmith/templates/listener/rbac.yaml
@@ -88,6 +88,13 @@ rules:
   - lgps/status
   verbs:
   - get
+- apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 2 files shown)


charts/langsmith/templates/listener/rbac.yaml
```diff
@@ -88,6 +88,13 @@ rules:
   - lgps/status
   verbs:
   - get
+- apiGroups:
```
langchain-infra: can we also update in the clusterrole?

langchain-infra (approved):

PR #532, opened 2025-12-04
mTLS support for Postgres + helm bump 0.12.28

romain-priour-lc:


charts/langsmith/templates/_helpers.tpl
```diff
@@ -167,7 +177,13 @@ Template containing common environment variables that are used by several servic
 - name: POSTGRES_SCHEMA
   value: {{ .Values.postgres.external.schema }}
 - name: POSTGRES_TLS
-  value: {{ .Values.postgres.external.customTls | quote }}
+  value: {{ or .Values.postgres.external.customTls (not (not .Values.postgres.external.clientCert.secretName)) | quote }}
```
romain-priour-lc: force TLS if client cert is passed to be safe

charts/langsmith/templates/_helpers.tpl
```diff
@@ -153,6 +153,16 @@ the user or some other secret provisioning mechanism
...
+{{/*
+Template for merging commonPodSecurityContext with component-specific podSecurityContext.
+Component-specific values take precedence over common values.
+Usage: {{ include "langsmith.podSecurityContext" (dict "Values" .Values "componentSecurityContext" .Values.backend.deployment.podSecurityContext) }}
+*/}}
+{{- define "langsmith.podSecurityContext" -}}
+{{- $merged := merge .componentSecurityContext .Values.commonPodSecurityContext -}}
+{{- toYaml $merged -}}
+{{- end -}}
+
```
romain-priour-lc: Should maybe add a validation that this is set if certs are passed?

langchain-infra (approved):

PR #546, opened 2025-12-26
chore: use backend service account for e2eTest

joaquin-borggio-lc:


charts/langsmith/templates/validate.yaml
```diff
@@ -164,3 +164,8 @@ It only performs validation checks and fails template rendering if issues are fo
 {{- if and .Values.clickhouse.external.clientCert.secretName (not .Values.clickhouse.external.existingSecretName) (not .Values.clickhouse.external.tls) -}}
 {{- fail "clickhouse.external.tls must be true when clickhouse.external.clientCert.secretName is set and the value is not being fetched from an existing secret." -}}
 {{- end -}}
+
+{{- /* Validate e2eTest serviceAccount removal */ -}}
+{{- if .Values.e2eTest.serviceAccount -}}
+{{- fail "e2eTest.serviceAccount is no longer supported. The e2eTest job now uses the backend service account. Please remove e2eTest.serviceAccount from your values." -}}
+{{- end -}}
```
joaquin-borggio-lc: adding this since its technically a backwards incompatible change. i dont think this will cause too many issues though

PR #544, opened 2025-12-24
chore: v13

joaquin-borggio-lc:

PR #548, opened 2026-01-06
v13: KEDA support for LangSmith queues + beacon logging + few issue fixes + deprecate basicauth initialOrgAdminEmail

romain-priour-lc:
- Not enabled by default.
- Gave users ample configuration
- Added validation and useful error messages if still using old configuration
- Also fix an issue where OTLP_ENDPOINT sticks around after disabling tracing, which ends up not disabling it.
- Fixed some comments around blob storage affecting README docs

Also added logs shipping to beacon, opt-out


langchain-infra (approved):

romain-priour-lc:
Closing, since merged into https://github.com/langchain-ai/helm/pull/544

PR #558, opened 2026-01-20
chore: add agent builder ci

joaquin-borggio-lc:


.github/workflows/release_charts.yaml
```diff
@@ -36,10 +36,11 @@ jobs:
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
           CR_SKIP_EXISTING: true
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
joaquin-borggio-lc: small thing to fix this warning:
```
Error while fetching latest Helm release: Error: [@octokit/auth-action] `GITHUB_TOKEN` variable is not set. It must be set on either `env:` or `with:`. See https://github.com/octokit/auth-action.js#createactionauth. Using default version v3.9.0
```
we already pass it in for CR_TOKEN.

PR #565, opened 2026-02-02
chore: allow disabling of agent products automatically

joaquin-borggio-lc:


charts/langsmith/templates/agent-bootstrap/bootstrap-job.yaml
```diff
@@ -1,9 +1,22 @@
...
+{{- if .Values.config.agentBuilder.enabled }}
+{{- $createProducts = append $createProducts "agent_builder" }}
+{{- else }}
+{{- $destroyProducts = append $destroyProducts "agent_builder" }}
+{{- end }}
+{{- if .Values.config.insights.enabled }}
+{{- $createProducts = append $createProducts "insights" }}
+{{- else }}
+{{- $destroyProducts = append $destroyProducts "insights" }}
+{{- end }}
```
romain-priour-lc: This logic would be better suited in helpers imo to not clog up the job definition

charts/langsmith/templates/agent-bootstrap/bootstrap-job.yaml
```diff
@@ -1,9 +1,22 @@
-{{- if and .Values.backend.agentBootstrap.enabled (or .Values.config.agentBuilder.enabled .Values.config.insights.enabled) }}
+{{- if and .Values.backend.agentBootstrap.enabled .Values.config.deployment.enabled }}
```
romain-priour-lc: IMO this agentboostrap.enabled variable is not necessary. The job is designed to be idempotent anyways. We should run the job if deployments and one of the agent features (insights or agentBuilder) is enabled. Is there value in letting users turn it off?

lmk what you think

charts/langsmith/templates/agent-bootstrap/bootstrap-job.yaml
joaquin-borggio-lc: fair. i can move these over

charts/langsmith/templates/agent-bootstrap/rbac.yaml
```diff
@@ -1,5 +1,5 @@
-# RBAC for the agent bootstrap job to create/patch ConfigMaps and restart deployments
-{{- if and .Values.backend.agentBootstrap.enabled (or .Values.config.agentBuilder.enabled .Values.config.insights.enabled) }}
+# ServiceAccount for the agent bootstrap job (needed whenever the job runs)
+{{- if and .Values.backend.agentBootstrap.enabled .Values.config.deployment.enabled }}
```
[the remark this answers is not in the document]
joaquin-borggio-lc: but same as above, we need to run this in the cleanup mode where both are disabled imo

romain-priour-lc (approved):

charts/langsmith/templates/agent-bootstrap/bootstrap-job.yaml
joaquin-borggio-lc: ok i decided to not keep it on by default just bc its changing a decent amount and im scared to cause issues for customers who dont even want agent products. we can change if needed later, but to avoid issues i want to keep default off

PR #575, opened 2026-02-09
fix(helm): address AKS/customer-reported issues in helm charts

cwaddingham:
## Summary

This PR applies four fixes reported by a customer deploying LangSmith and Langgraph Dataplane on self-hosted AKS with Helm. The changes touch both **langsmith** and **langgraph-dataplane** charts and address operator pod spec behavior, template indentation, RBAC naming, and configurable Ingress namespace.

---

### 1. `automountServiceAccountToken: false` not applied on operator pod

**Charts:** Langsmith, Langgraph Dataplane  
**Template:** `operator/deployment.yaml` (both charts)

The `automountServiceAccountToken: false` setting was not taking effect for the operator deployment. The value was only applied on the ServiceAccount resource; in environments like AKS (e.g. with workload identity), the **pod spec** must set `automountServiceAccountToken` for it to be respected.

### 2. Indentation error in volume mount (Langgraph Dataplane)

**Chart:** Langgraph Dataplane  
**Template:** `langgraph-dataplane/templates/operator/deployment.yaml`

Extra volume mounts from values were rendered with incorrect indentation. The block used `nindent 12`, which over-indented the appended volume mounts relative to the existing `volumeMounts` list (10 spaces), producing invalid or misaligned YAML.

### 3. Role and RoleBinding names when using an existing service account

**Charts:** Langsmith, Langgraph Dataplane  
**Templates:** `listener/rbac.yaml`, `operator/rbac.yaml` (both charts)

Role and RoleBinding names were derived from the service account name (e.g. `{{ include "operator.serviceAccountName" $ }}-role`). When using an **existing** service account (e.g. for AKS workload identity), that name does not reflect the Helm release; the customer needed names that identify the release/component (e.g. release-name-listener, release-name-operator). Using the SA name also risks **name collisions** when multiple releases share the same existing SA in the same namespace.

### 4. No option to set Ingress resource namespace (Langgraph Dataplane)

**Chart:** Langgraph Dataplane  
**Template:** `langgraph-dataplane/templates/ingress.yaml`

The Ingress resource was always created in the release namespace. There was no way to create it in another namespace (e.g. where the cluster’s ingress controller or existing Ingress lives). The operator’s `SHARED_INGRESS_NAMESPACE` env was also tied to the release namespace, so it could not point at an Ingress in a different namespace.

---

### 1. `automountServiceAccountToken` on pod spec

- **Langsmith:** In `charts/langsmith/templates/operator/deployment.yaml`, added `automountServiceAccountToken: {{ .Values.operator.serviceAccount.automountServiceAccountToken | default true }}` under the pod `spec` (next to `serviceAccountName`).
- **Langgraph Dataplane:** Same addition in `charts/langgraph-dataplane/templates/operator/deployment.yaml`.

The existing `operator.serviceAccount.automountServiceAccountToken` value now applies at the pod level; default remains `true` for backward compatibility.

### 2. Volume mount indentation

- In `charts/langgraph-dataplane/templates/operator/deployment.yaml`, changed the extra volume mounts block from `nindent 12` to `nindent 10` so they align with the existing `volumeMounts` list.

### 3. Role/RoleBinding names from release fullname

- **Langsmith**
  - **Listener** (`charts/langsmith/templates/listener/rbac.yaml`): Role name, RoleBinding name, and `roleRef.name` now use `{{ include "langsmith.fullname" . }}-listener` instead of `listener.serviceAccountName`-based names.
  - **Operator** (`charts/langsmith/templates/operator/rbac.yaml`): All operator Role/ClusterRole and RoleBinding/ClusterRoleBinding names (and their `roleRef`s) use fullname-based names: `...-operator` for the main RBAC and `...-operator-leader-election-role` / `...-operator-leader-election-rolebinding` for leader election. In the `watchNamespaces` range, `{{ include "langsmith.fullname" $ }}` is used; in the cluster-wide and leader-election blocks, `{{ include "langsmith.fullname" . }}` is used.
- **Langgraph Dataplane**
  - **Listener** (`charts/langgraph-dataplane/templates/listener/rbac.yaml`): Same pattern with `{{ include "langgraphDataplane.fullname" $ }}-listener` for Role, RoleBinding, and `roleRef.name`.
  - **Operator** (`charts/langgraph-dataplane/templates/operator/rbac.yaml`): Same pattern with `langgraphDataplane.fullname` for operator and leader-election Role/RoleBinding names.

`subjects` still reference the (existing or chart-created) service account by name; only the **names** of the Role and RoleBinding resources were changed so they are unique per release and reflect the component.

### 4. Optional Ingress namespace (Langgraph Dataplane)

- **`charts/langgraph-dataplane/values.yaml`:** Under `ingress:`, added `namespace: ""` with a comment that it is the namespace where the Ingress is created and where the operator will look for it (defaults to release namespace when unset).
- **`charts/langgraph-dataplane/templates/ingress.yaml`:** Ingress `metadata.namespace` is now `{{ .Values.ingress.namespace | default .Values.namespace | default .Release.Namespace }}`.
- **`charts/langgraph-dataplane/templates/operator/deployment.yaml`:** When ingress is enabled, `SHARED_INGRESS_NAMESPACE` uses the same expression so the operator looks for the Ingress in the configured namespace.

When `ingress.namespace` is not set, behavior is unchanged. When set (e.g. `ingress.namespace: "ingress-nginx"`), the Ingress is created in that namespace and the operator is configured to use it.

---

## Testing Performed

- **Helm template rendering**
  - **Langsmith:** `helm template` was run with deployment enabled, ingress enabled, operator and RBAC enabled, and required config (e.g. `config.hostname`, license/key/auth) to confirm templates render without errors. Verified that the operator deployment pod spec includes `automountServiceAccountToken` and that Role/RoleBinding names are fullname-based (e.g. `release-name-langsmith-operator`, `release-name-langsmith-listener`).
  - **Langgraph Dataplane:** `helm template` was run with ingress and operator enabled (and related settings). Verified:
    - Operator pod spec includes `automountServiceAccountToken`.
    - Role/RoleBinding names are fullname-based (e.g. `release-name-langgraph-dataplane-operator`, `release-name-langgraph-dataplane-listener`).
    - With default values, Ingress and `SHARED_INGRESS_NAMESPACE` use the release namespace.
    - With `ingress.namespace=ingress-nginx`, the rendered Ingress has `namespace: ingress-nginx` and the operator env has `SHARED_INGRESS_NAMESPACE=ingress-nginx`.
- **Syntax:** No new linter errors were introduced for the modified templates. Rendered YAML structure and indentation (including the fixed volume mounts) were spot-checked in the template output.

### DIFF
--- a/charts/langgraph-dataplane/Chart.yaml
+++ b/charts/langgraph-dataplane/Chart.yaml
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy a langgraph dataplane on kubernetes.
 type: application
-version: 0.2.16
+version: 0.2.17
 appVersion: "0.13.9"
--- a/charts/langsmith/Chart.yaml
+++ b/charts/langsmith/Chart.yaml
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
-version: 0.13.9
+version: 0.13.10
 appVersion: "0.13.9"


charts/langgraph-dataplane/templates/listener/rbac.yaml
```diff
@@ -10,7 +10,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: {{ include "listener.serviceAccountName" $ }}-role
+  name: {{ include "langgraphDataplane.fullname" $ }}-listener
```
langchain-infra: is there a reason for this change? we do this for all service accounts to allow people to override the name in case they need to make the service accounts themselves.

charts/langgraph-dataplane/templates/operator/deployment.yaml
```diff
@@ -121,16 +122,12 @@ spec:
           - name: SHARED_INGRESS_NAME
             value: {{ include "langgraphDataplane.fullname" . }}-ingress
           - name: SHARED_INGRESS_NAMESPACE
-            value: {{ .Values.namespace | default .Release.Namespace }}
+            value: {{ .Values.ingress.namespace | default .Values.namespace | default .Release.Namespace }}
```
langchain-infra: does this work? AFAIK you can't have ingress resources in a different namespace from the service they point to

langchain-infra (approved):
may want to bump chart versions

PR #593, opened 2026-02-25
chore: remove marketplace script, refactor mirror script

jkennedyvz:
## Summary
- Remove `create_marketplace_version.sh` (moved to deployments repo)
- Refactor `mirror_langsmith_images.sh` to:
  - Read image list from `values.yaml` instead of hardcoding `DEFAULT_VERSION` — images always match the chart
  - Remove `--dest-repo` marketplace mode (AWS-specific, moved to deployments)
  - Keep `--version` as optional override for LangChain images (infra image tags stay as-is)

## Addresses review feedback
- @langchain-infra: keep the mirror script, remove only AWS-specific components
- @romain-priour-lc: read versions from values.yaml to avoid hardcoded values
- @romain-priour-lc: no values.yaml version bumps in this PR

## Test plan
- [x] `--dry-run` correctly reads all 14 images from values.yaml
- [x] `--version` override applies only to langchain images, not postgres/redis/clickhouse
- [x] `create_marketplace_version.sh` removed


charts/langsmith/scripts/MARKETPLACE_RELEASE.md
```diff
@@ -0,0 +1,140 @@
...
+| `--version` | `0.13.14` | LangSmith image version |
+| `--operator-version` | `0.1.37` | langgraph-operator version |
+| `--chart-version` | same as `--version` | Helm chart tag in ECR |
+| `--platform` | `linux/amd64` | Image architecture to pull |
+
+## Full example with overrides
+
+```bash
+VERSION=0.13.15
+OPERATOR_VERSION=0.1.38
```
romain-priour-lc: Will we ever need to push older versions? I wonder if we should instead just parse values.yaml for the current chart version. This seems a little error prone

charts/langsmith/scripts/mirror_langsmith_images.sh
langchain-infra: we still want this, we just odnt need the aws specific components

PR #665, opened 2026-04-02
feat(langgraph-cloud): add Gateway API and Istio VirtualService support

saad-supports-langchain:
Add support for Kubernetes Gateway API (HTTPRoute) and Istio (VirtualService) as ingress alternatives alongside the existing Ingress resource. Only one mode can be enabled at a time, enforced by a mutual exclusivity validation.

- Add gateway and istioGateway sections to values.yaml
- Add http_route.yaml template with basePath + URLRewrite support
- Add virtual_service.yaml template with basePath rewrite (exact + prefix rules)
- Service templates force ClusterIP and drop HTTPS port when mesh ingress is active
- Add validation for required fields (gateway.name, istioGateway.name, ingress.hostname)
- Guard ingressClassName and loadBalancerIP against empty values
- Normalize basePath with trimPrefix/trimSuffix to handle leading/trailing slashes
- Match langsmith chart conventions (file naming, namespace formatting)


hari-dhanushkodi (approved):
lets bump the minor version of this helm chart, and update docs using `helm-docs -t ./charts/langgraph-cloud/README.md.gotmpl`. I think after this, we should cut a breaking release and update major version that removes standalone studio

charts/langgraph-cloud/values.yaml
```diff
@@ -76,6 +76,44 @@ ingress:
...
+istioGateway:
+  enabled: false
+  # -- Name of the Istio Gateway resource.
+  name: ""
+  # -- Namespace of the Istio Gateway resource.
+  namespace: ""
+  # -- Hostname for the VirtualService. If not set, the VirtualService will match all hosts ("*").
+  hostname: ""
+  # -- Optional hostname for studio VirtualService. If not set, studio routes are included on the main hostname.
+  studioHostname: ""
```
langchain-infra: We should just drop studio. The debugger image doesn't get built anymore

hari-dhanushkodi: i think it does lol: https://hub.docker.com/r/langchain/langgraph-debugger/tags, but ya standalone studio we can drop

saad-supports-langchain: studio dropped in latest commit, also bumped chart version to 0.3.0

langchain-infra (approved):

PR #693, opened 2026-04-27
feat: in chart standalone agents

joaquin-borggio-lc:


charts/langsmith/ci/standalone-agents-values.yaml
```diff
@@ -0,0 +1,202 @@
+# Standalone agent features configuration.
```
langchain-infra: nit, now should we just include this in like the readonly chart or something? i dont think it needs to be a separate job.

langchain-infra (approved):
i kinda wonder if we should just keep redis separate and then do the consolidation separately. users can still combine this if they want like pg. Keeps it the same as pg. Mainly just because we are moving some operations from pg -> redis
i think we can add the agent features to the main readonly and regular ci?

PR #703, opened 2026-05-05
chore: support byoc deployments

joaquin-borggio-lc:


charts/langsmith/templates/virtual_service.yaml
romain-priour-lc, line 1: nit, should we add frontend check to ingress and gateway as well?

romain-priour-lc (approved):

charts/langsmith/templates/virtual_service.yaml
joaquin-borggio-lc: rn if you enable ingress or gateway without frontend itll error. only istio allows no frontend. i think fine to leave to keep guardrails tight. later we could allow more combos if we want. gonna keep off for now

PR #706, opened 2026-05-06
chore: use api key salt, not jwt secret

joaquin-borggio-lc:


langchain-infra (approved):
why do we have this separate set of templated values? it leads to errors like this. would like to potentially consolidate with our main templates

joaquin-borggio-lc:
> why do we have this separate set of templated values? it leads to errors like this. would like to potentially consolidate with our main templates

we dont mount every secret/env var like we do for other services. we could tests but in saas we dont do that so would like to test. i can move these helpers to that file. ill do in follow up

PR #709, opened 2026-05-08
fix: chop names even more for hash thats added

joaquin-borggio-lc:


charts/langsmith/templates/agent-features/_helpers.tpl
```diff
@@ -10,6 +10,53 @@ Produces: <release>-<namePrefix>  e.g. "langsmith-standalone-fleet"
 {{- printf "%s-%s" (include "langsmith.fullname" $root) $prefix | trunc 63 | trimSuffix "-" }}
 {{- end -}}
 
+{{/*
```
joaquin-borggio-lc, line 13: i have a todo to combine this with the other helpers tpl which i have a pr for but needs more qa so leaving in this one for now while there are two

romain-priour-lc (approved):

PR #714, opened 2026-05-11
feat(langsmith): add preInstallManifests for pre-install hook ordering

karantyagi-reltio:
## Summary

Adds a top-level `preInstallManifests` list to the `langsmith` chart. Each entry is rendered as a Helm `pre-install` / `pre-upgrade` hook so the resource is applied **before** the rest of the chart's manifests.

Intended for prerequisites the chart consumes during install/upgrade — most commonly an `ExternalSecret` (External Secrets Operator) whose materialized Kubernetes Secret is referenced via `config.existingSecretName` or the per-datastore `*.external.existingSecretName` values.

This is scoped intentionally narrow: arbitrary side-car manifests are not covered here — those are better handled by Helm composition (parent chart with `langsmith` as a dependency). The gap composition does **not** cover is ordering: subchart and parent manifests render into a single release manifest list and Helm sorts by `kind`, not by chart, so an ExternalSecret declared in a parent chart is not guaranteed to be applied before the `langsmith` Deployments that consume it.

## Usage

```yaml
preInstallManifests:
  - apiVersion: external-secrets.io/v1beta1
    kind: ExternalSecret
    metadata:
      name: langsmith-app
    spec:
      refreshInterval: 1h
      secretStoreRef:
        name: vault-backend
        kind: ClusterSecretStore
      target:
        name: langsmith-app-secret      # matches config.existingSecretName
        creationPolicy: Orphan          # so the materialized Secret survives upgrades
      data:
        - secretKey: langsmith_license_key
          remoteRef:
            key: secret/langsmith/app
            property: langsmith_license_key
```

The chart then consumes the materialized Secret via the existing knobs:

```yaml
config:
  disableSecretCreation: true
  existingSecretName: langsmith-app-secret
```

Entries can also be multi-line YAML strings; both forms run through `tpl`, so `{{ .Release.Name }}` etc. works.

## Annotation handling (idempotent)

The following annotations are injected automatically:

```
helm.sh/hook: pre-install,pre-upgrade
helm.sh/hook-weight: "-5"
helm.sh/hook-delete-policy: before-hook-creation
```

User-provided values on the same keys take precedence, so manifests that already carry hook annotations are handled correctly — partial overrides are merged (any default the user did not set is still applied). Example: a user setting `helm.sh/hook-weight: "-50"` keeps that weight while still receiving the default `hook` and `hook-delete-policy` values.

## Implementation

- `templates/pre-install-manifests.yaml` — iterates `.Values.preInstallManifests`, renders each entry through `tpl`, then `merge`s user annotations over the default hook annotations (user wins). Header comment documents entry forms, annotation semantics, idempotency, and the ExternalSecret `creationPolicy: Orphan` caveat in detail.
- `values.yaml` — short comment + simple `ExternalSecret` example; defers to the template file for full details.

Default is `[]`, so behavior is unchanged for existing users.

## Caveats

Hook resources are not tracked as part of the Helm release and are recreated on every upgrade. For `ExternalSecret` entries, prefer `target.creationPolicy: Orphan` (or `target.deletionPolicy: Retain` on newer ESO versions) so the materialized Secret survives upgrades and pods consuming it don't restart.

Each entry must be a single resource (no `---` separators inside one string).

---

Scoped down from the original PR per review feedback in #714. `extraManifests` was dropped — composition via subchart is the recommended pattern for arbitrary side-car resources.

### DIFF
--- a/charts/langsmith/templates/extra-manifests.yaml
+++ b/charts/langsmith/templates/extra-manifests.yaml
@@ -1,0 +1,8 @@
+{{- range .Values.extraManifests }}
+---
+{{- if typeIs "string" . }}
+{{ tpl . $ }}
+{{- else }}
+{{ tpl (toYaml .) $ }}
+{{- end }}
+{{- end }}
--- a/charts/langsmith/templates/extra-manifests.yaml
+++ b/charts/langsmith/templates/extra-manifests.yaml
@@ -1,8 +1,0 @@
-{{- range .Values.extraManifests }}
----
-{{- if typeIs "string" . }}
-{{ tpl . $ }}
-{{- else }}
-{{ tpl (toYaml .) $ }}
-{{- end }}
-{{- end }}
--- a/charts/langsmith/templates/pre-install-manifests.yaml
+++ b/charts/langsmith/templates/pre-install-manifests.yaml
@@ -1,18 +1,35 @@
 {{- /*
-Renders each entry in .Values.preInstallManifests as a Helm pre-install / pre-upgrade
-hook so the resource is applied before the rest of the chart's manifests. Useful for
-prerequisites the chart consumes during install/upgrade (for example, an
-ExternalSecret whose materialized Secret is referenced by the chart's pods).
+Renders each entry in .Values.preInstallManifests as a Helm pre-install /
+pre-upgrade hook so the resource is applied before the rest of the chart's
+manifests. Intended for prerequisites the chart consumes during install/upgrade
+-- for example, an ExternalSecret (External Secrets Operator) whose
+materialized Kubernetes Secret is referenced via `config.existingSecretName`
+or the per-datastore `*.external.existingSecretName` values.
 
-Each entry can be a YAML object or a multi-line YAML string. Both are rendered
-through `tpl` so template expressions work. Each entry must be a single resource
-(no `---` separators inside one string).
+Entry forms
+-----------
+Each entry may be a YAML object or a multi-line YAML string. Both are rendered
+through `tpl` so template expressions like `{{ .Release.Name }}` work. Each
+entry must be a single resource (no `---` separators inside one string).
 
-The following annotations are injected automatically; user-provided annotations
-on the same keys win, so they can be overridden per entry:
+Annotation handling
+-------------------
+The following annotations are injected automatically. User-provided values for
+the same keys take precedence, so the template is idempotent for manifests
+that already carry hook annotations -- partial overrides are merged
+(any default the user did not set is still applied).
+
   helm.sh/hook: pre-install,pre-upgrade
   helm.sh/hook-weight: "-5"
   helm.sh/hook-delete-policy: before-hook-creation
+
+Caveats
+-------
+Hook resources are not tracked as part of the Helm release and are recreated
+on every upgrade. For ExternalSecret entries, prefer
+`target.creationPolicy: Orphan` (or `target.deletionPolicy: Retain` on newer
+ESO versions) so the materialized Secret survives upgrades and pods consuming
+it don't restart.
 */ -}}
 {{- range .Values.preInstallManifests }}
 {{- $manifest := . }}
--- a/charts/langsmith/templates/pre-install-manifests.yaml
+++ b/charts/langsmith/templates/pre-install-manifests.yaml
@@ -1,0 +1,33 @@
+{{- /*
+Renders each entry in .Values.preInstallManifests as a Helm pre-install / pre-upgrade
+hook so the resource is applied before the rest of the chart's manifests. Useful for
+prerequisites the chart consumes during install/upgrade (for example, an
+ExternalSecret whose materialized Secret is referenced by the chart's pods).
+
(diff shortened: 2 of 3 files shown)


joaquin-borggio-lc:
For adding arbitrary manifests, I would recommend using helm composition.

something where you have this kind of structure:
```
my-langsmith/
  Chart.yaml
  values.yaml
  templates/
    my-configmap.yaml
```

this is your chart.yaml:
```
# Chart.yaml
apiVersion: v2
name: my-langsmith
version: 0.1.0
dependencies:
  - name: langsmith
    version: "0.x.x"
    repository: "https://langchain-ai.github.io/helm"
```

this anything you want in the templates folder. example:
```
# templates/my-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}-custom-config
  namespace: {{ .Release.Namespace }}
data:
  APP_ENV: production
  SOME_FLAG: "true"
```

and your values file would look like this:
```
# values.yaml
langsmith:
  config:
    existingSecretName: langsmith-app-secret
  # any other langsmith values nested here
```

That's the general idea with the helm composition that allows you to combine different manifests or helm charts.

karantyagi-reltio:
Thanks @joaquin-borggio-lc — composition is the right tool for arbitrary side-car manifests, and I'm convinced it covers the `extraManifests` half of this PR.

The reason I'd still like to land the `preInstallManifests` half is **install/upgrade ordering**, which composition genuinely doesn't solve:

In a parent/child chart setup, both layers render into a single release manifest list, and Helm sorts by `kind`, not by chart. So a parent-chart `ExternalSecret` is **not** guaranteed to be applied (and reconciled by ESO into the target `Secret`) before the `langsmith` Deployments that consume it. On a clean install you get a window where the pods reference a not-yet-materialized Secret and crash-loop until ESO catches up; on upgrade you get a brief flap whenever the ES spec changes.

The standard fix is `helm.sh/hook: pre-install,pre-upgrade` with a negative weight, which is what `preInstallManifests` injects. You *can* do this by hand in a wrapper chart by annotating every entry, but:

- it's easy to forget on one resource and silently get pod restarts,
- the annotation set (`hook`, `hook-weight`, `hook-delete-policy`) and its caveats (e.g. `creationPolicy: Orphan` so the materialized Secret survives upgrades) are non-obvious,
- ESO + LangSmith is a common enough combination that having a blessed path in the chart saves every adopter the same debug cycle.

karantyagi-reltio:
@joaquin-borggio-lc Any update here ?

CC: @saad-supports-langchain

karantyagi-reltio:
@joaquin-borggio-lc Any update here ?

joaquin-borggio-lc:
@karantyagi-reltio  I see your point of potentially forgetting the annotations. I think the following then makes sense since extra manifests can be achieved with composition:
1. Scope this PR to `preInstallManifests` only to handle the ordering. Let's make sure we can handle a manifest passed in with the annotations already added as well in case someone uses this with the annotations.
2. if we could also reduce the size of the comments in the values.yaml. That file already is very long and hard to parse. Let's keep the comment above `preInstallManifests` to be a couple lines of explanation and a simple example. Any additional info I'd put in `pre-install-manifests.yaml`

karantyagi-reltio:
Thanks @joaquin-borggio-lc — all three points addressed in [1d3857f](https://github.com/langchain-ai/helm/pull/714/commits/1d3857f):

1. **Scoped to `preInstallManifests` only.** `extraManifests` and `templates/extra-manifests.yaml` are removed; composition handles that path. Title and PR body updated to match.
2. **Manifests with hook annotations already added are handled.** The template `merge`s user-supplied annotations over the injected defaults, so user values win on conflict. Partial overrides also work — any default the user did not set is still applied (e.g. a user can set `helm.sh/hook-weight: "-50"` and still receive the default `hook` / `hook-delete-policy`). Verified with three cases (no annotations, full override, partial override) — all render as expected.
3. **`values.yaml` comment slimmed.** Now ~5 lines of explanation + a simple `ExternalSecret` example. Full details (entry forms, injected annotations, idempotency rules, `creationPolicy: Orphan` caveat) live in the header comment of `templates/pre-install-manifests.yaml`.

Let me know if you'd like any further tweaks.

joaquin-borggio-lc (approved):

karantyagi-reltio:
@joaquin-borggio-lc 

[Lint and Test Charts / lint-test (langsmith) (pull_request)](https://github.com/langchain-ai/helm/actions/runs/28089363123/job/83213467273?pr=714) are failing as i have forked this repository and GitHub Actions does not pass repository secrets to PRs from forks so for license it is failing.

karantyagi-reltio:
How we can handle and solve this ? Please help here

joaquin-borggio-lc:
merged it here: https://github.com/langchain-ai/helm/pull/780

joaquin-borggio-lc:
feel free to close out this pr. the functionality is merged into main now

karantyagi-reltio:
sure thank you what is the latest chart version where i can use this ?

PR #716, opened 2026-05-11
chore: move agent features to shared config map and helper.tpl

joaquin-borggio-lc:


charts/langsmith/templates/_helpers.tpl
```diff
@@ -658,6 +658,177 @@ Template containing common environment variables that are used by several servic
...
+Extra env vars for insights api-server and queue pods.
+*/}}
+{{- define "langsmith.insights.extraEnv" -}}
+{{- $root := index . "root" -}}
+{{- $componentName := index . "component" -}}
+{{- $feature := $root.Values.insights -}}
+{{- $component := index $feature $componentName -}}
+{{- $ns := $root.Values.namespace | default $root.Release.Namespace -}}
+{{- $cd := $root.Values.clusterDomain -}}
+{{- $platformBackend := printf "http://%s-%s.%s.svc.%s:%v" (include "langsmith.fullname" $root) $root.Values.platformBackend.name $ns $cd $root.Values.platformBackend.service.port -}}
```
langchain-infra: why add this as extra env?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -658,6 +658,177 @@ Template containing common environment variables that are used by several servic
...
+  (dict "name" "POSTGRES_URI" "valueFrom" (dict "secretKeyRef" (dict "name" (include "langsmith.agentFeatures.postgresSecretName" (dict "root" $root "product" "fleet")) "key" "postgres_connection_url")))
+  (dict "name" "REDIS_URI" "valueFrom" (dict "secretKeyRef" (dict "name" (include "langsmith.agentFeatures.redisSecretName" (dict "root" $root "product" "fleet")) "key" "redis_connection_url")))
+  (dict "name" "LANGSMITH_AUTH_ENDPOINT" "value" $platformBackend)
+  (dict "name" "MCP_SERVER_URL" "value" $toolServer)
+  (dict "name" "LANGSMITH_LICENSE_REQUIRED_CLAIMS" "value" "agent_builder_enabled")
+  (dict "name" "SSRF_ALLOW_PRIVATE_IPS_MCP_SERVERS" "value" "true")
+  (dict "name" "SSRF_ALLOW_PRIVATE_IPS_TOOLS" "value" "true")
+  (dict "name" "SSRF_ALLOW_K8S_INTERNAL" "value" "true")
+-}}
+{{- if and (eq $componentName "apiServer") $feature.queue.enabled -}}
```
langchain-infra, line 759: we should just always deploy with queue

charts/langsmith/templates/_helpers.tpl
```diff
@@ -658,6 +658,177 @@ Template containing common environment variables that are used by several servic
...
+*/}}
+{{- define "langsmith.fleet.extraEnv" -}}
+{{- $root := index . "root" -}}
+{{- $componentName := index . "component" -}}
+{{- $feature := $root.Values.fleet -}}
+{{- $component := index $feature $componentName -}}
+{{- $ns := $root.Values.namespace | default $root.Release.Namespace -}}
+{{- $cd := $root.Values.clusterDomain -}}
+{{- $platformBackend := printf "http://%s-%s.%s.svc.%s:%v" (include "langsmith.fullname" $root) $root.Values.platformBackend.name $ns $cd $root.Values.platformBackend.service.port -}}
+{{- $toolServer := printf "http://%s-%s.%s.svc.%s:%v" (include "langsmith.fullname" $root) $root.Values.agentBuilderToolServer.name $ns $cd $root.Values.agentBuilderToolServer.service.port -}}
```
langchain-infra: wouldnt these just be included in our configmap?

langchain-infra (approved):

charts/langsmith/templates/_helpers.tpl
joaquin-borggio-lc: we need LANGSMITH_AUTH_ENDPOINT to point to platform backend. i could add it to commonEnv but would want to test that doesnt break anything (i doubt it)

charts/langsmith/templates/_helpers.tpl
joaquin-borggio-lc: probably. maybe a customer wants them combined if they have some kinda of issue with ips or pods idk. i can do in a follow up to not have this pr get too big and cross scope but i probably agree we should avoid them deploying without queue

charts/langsmith/templates/_helpers.tpl
joaquin-borggio-lc: only fleet needs them, but i can move it to config map. REIDS_URI nad POSTGRES_URI are not in our commonEnv. those are actually REDIS_DATABASE_URI and POSTGRES_DATABASE_URI. but we need them to be diff bc we dont want to use the LS databases

joaquin-borggio-lc: so if anything we could move MCP_SERVER_URL and LANGSMITH_AUTH_ENDPOINT to our config map. but the others id keep here

PR #717, opened 2026-05-12
feat: add extraPorts support to queue deployment

simon-langchain:
## Summary

- Adds `queue.deployment.extraPorts: []` to `values.yaml`
- Adds a `{{- with .Values.queue.deployment.extraPorts }}` block to the queue deployment template, appending any extra ports after the primary health/metrics port

## Context

The queue deployment currently exposes a single named container port (the internal health+metrics server on `containerPort`). Customers who run custom metrics endpoints inside their graph code need to expose an additional named port so that Prometheus Operator's PodMonitor can scrape it — `targetPort` is deprecated in PodMonitors in favor of named ports.

## Usage

```yaml
queue:
  deployment:
    extraPorts:
      - name: custom-metrics
        containerPort: 9090
        protocol: TCP
```

## Test plan

- [ ] `helm template` renders the extra port when `extraPorts` is set
- [ ] `helm template` renders cleanly (no extra ports block) when `extraPorts` is empty

### DIFF
--- a/charts/langgraph-cloud/templates/queue/deployment.yaml
+++ b/charts/langgraph-cloud/templates/queue/deployment.yaml
@@ -109,6 +109,9 @@ spec:
             - name: queue
               containerPort: {{ .Values.queue.containerPort }}
               protocol: TCP
+            {{- with .Values.queue.deployment.extraPorts }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
           startupProbe:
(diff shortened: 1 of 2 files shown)


saad-supports-langchain:
screenshot



works

saad-supports-langchain (approved):

joaquin-borggio-lc (approved):

PR #735, opened 2026-05-26
Cherry-pick #729 + #733 into v15-release

rx5ad:
## Summary
Cherry-picks the following PRs from `main` into `v15-release`:

- **#729** — `feat: add nginx routes for /api/v1/fleet/ to platformBackend` (clean apply)
- **#733** — `fix(fleet): inject LANGGRAPH_DEPLOYMENT_URL into platform-backend when …` (took only the `_helpers.tpl` change; dropped the incidental Chart.yaml version bump and helm-docs README regeneration that conflicted with this branch's release line)

**#730** (`chore(langsmith): bump chart to 0.15.0-rc.16`) was intentionally skipped — `v15-release` is on its own version line (`0.15.0-rc.15` / appVersion `0.15.8rc1`) and #730 would have regressed `appVersion` from `0.15.8rc1` to `0.15.7rc1`.

Net diff vs `v15-release`:
- `charts/langsmith/templates/_helpers.tpl` (+7)
- `charts/langsmith/templates/frontend/config-map.yaml` (+18)

## Test plan
- [ ] Verify nginx routes for `/api/v1/fleet/` resolve in a v15 deploy
- [ ] Verify `LANGGRAPH_DEPLOYMENT_URL` is injected into platform-backend when fleet is enabled
- [ ] Confirm chart version (`0.15.0-rc.15`) and appVersion (`0.15.8rc1`) are unchanged on this branch

### DIFF
--- a/charts/langsmith/templates/_helpers.tpl
+++ b/charts/langsmith/templates/_helpers.tpl
@@ -201,6 +201,13 @@ Template containing common environment variables that are used by several servic
 - name: HOST_BACKEND_ENDPOINT_PUBLIC
   value: {{ .Values.config.hostname }}/api-host
 {{- end }}
+{{- if .Values.fleet.enabled }}
+{{- $ns := .Values.namespace | default .Release.Namespace -}}
+{{- $cd := .Values.clusterDomain -}}
+{{- $fleetApi := printf "http://%s.%s.svc.%s:%v" (include "langsmith.agentFeatures.apiServerK8sServiceName" (dict "root" . "product" "fleet")) $ns $cd .Values.fleet.apiServer.service.httpPort }}
+- name: LANGGRAPH_DEPLOYMENT_URL
+  value: {{ $fleetApi | quote }}
+{{- end }}
 - name: REDIS_CLUSTER_ENABLED
   value: {{ .Values.redis.external.cluster.enabled | quote }}
 {{- if .Values.redis.external.cluster.enabled }}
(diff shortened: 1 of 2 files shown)


charts/langsmith/templates/_helpers.tpl
```diff
@@ -201,6 +201,13 @@ Template containing common environment variables that are used by several servic
 - name: HOST_BACKEND_ENDPOINT_PUBLIC
   value: {{ .Values.config.hostname }}/api-host
 {{- end }}
+{{- if .Values.fleet.enabled }}
```
langchain-infra, line 204: maybe two nits (one prob requires a lcplus change), should this be `FLEET_URL`? also can we put int he config-map with the other endpoint urls.

rx5ad: You mean `FLEET_URL_LANGGRAPH_DEPLOYMENT_URL`?

I was thinking about it but decided against it since the URLs belong to upstream services, and it would add duplication. Also, it would introduce yet another convention.

Is it okay if we do this separately?

langchain-infra: yea can be a separate change. but we can also have many more upstream LANGSMITH_DEPLOYMENT services (e.g doesn't demarcate from insights). Generally we title this by the upstream service (so would be falling the convention). Auth service, playground service etc, how would it be breaking convention?

rx5ad: Fair. By convention I meant we don't use FLEET_URL_ prefix for other upstream dependencies of smith-go.

PR #736, opened 2026-05-27
feat(smithdb): Preview In-Chart SmithDB

ben11211:
## Description

Disabled by default SmithDB chart configuration.


charts/langsmith/templates/smithdb/client-deployment.yaml
langchain-infra, line 1: nit, do we need this?

charts/langsmith/templates/smithdb/ingestion-deployment.yaml
```diff
@@ -0,0 +1,115 @@
+{{- if and .Values.smithdb.enabled .Values.smithdb.ingestion.enabled }}
+{{- $clusterManagerEnv := list -}}
+{{- if .Values.smithdb.clusterManager.enabled }}
+{{- $clusterManagerEnv = (include "langsmith.smithdb.clusterManagerClientEnv" (dict "root" . "service" "INGESTION") | fromYamlArray) -}}
```
langchain-infra: why do we need clusterManagerEnv here? the env management is a bit confusing. Also do we need to add commonEnv?

charts/langsmith/templates/smithdb/pdbs.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if .Values.smithdb.enabled }}
+{{- range $component := list "query" "ingestion" }}
```
langchain-infra: why not support pdbs on all components? insteead of just query/ingestion?

charts/langsmith/templates/smithdb/service-accounts.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if .Values.smithdb.enabled }}
+{{- range $component := list "query" "ingestion" "compaction" "client" "clusterManager" "metastoreMigration" }}
```
langchain-infra: this is mainly a nit, i prefer adding separate files for stuff like this. Separately, should we just put all these under one smithdb service account? Like we do for smith-backend and all jobs associated with it.

charts/langsmith/templates/_helpers.tpl
```diff
@@ -485,6 +485,339 @@ Template containing common environment variables that are used by several servic
...
+{{/*
+SmithDB HTTP-style endpoint used by SmithDB services.
+Args: root, component, port.
+*/}}
+{{- define "langsmith.smithdb.httpServiceUrl" -}}
+{{- printf "http://%s" (include "langsmith.smithdb.grpcServiceUrl" .) }}
+{{- end }}
+
+{{/*
+SmithDB HTTP health probes.
```
langchain-infra: should we allow passing these in like other sections?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -485,6 +485,339 @@ Template containing common environment variables that are used by several servic
...
+    port: http
+  failureThreshold: 6
+  periodSeconds: 10
+  timeoutSeconds: 1
+{{- end }}
+
+{{/*
+SmithDB gRPC health probes. Args: port.
+*/}}
+{{- define "langsmith.smithdb.grpcProbes" -}}
```
langchain-infra: same here

charts/langsmith/templates/_helpers.tpl
```diff
@@ -485,6 +485,339 @@ Template containing common environment variables that are used by several servic
...
+readinessProbe:
+  grpc:
+    port: {{ .port }}
+  failureThreshold: 6
+  periodSeconds: 10
+  timeoutSeconds: 1
+{{- end }}
+
+{{/*
+SmithDB cluster-manager HTTP endpoint used by SmithDB services.
```
langchain-infra, line 541: why add this? should this be an env var we pass around instead? its how our other service urls are defined in the main config-map.yaml

charts/langsmith/templates/_helpers.tpl
```diff
@@ -485,6 +485,339 @@ Template containing common environment variables that are used by several servic
...
+- name: {{ $prefix }}__STATUS_BUFFER_SIZE
+  value: "32"
+- name: {{ $prefix }}__CONNECT_TIMEOUT
+  value: "1s"
+{{- end }}
+
+{{/*
+SmithDB metastore migration env vars.
+*/}}
+{{- define "langsmith.smithdb.metastoreMigrationEnv" -}}
```
langchain-infra: should we just inline these to the component deployment?

charts/langsmith/values.yaml
```diff
@@ -101,6 +101,30 @@ images:
     repository: "docker.io/langchain/langsmith-polly"
     pullPolicy: IfNotPresent
     tag: "0.15.11"
+  smithdbQueryImage:
```
langchain-infra, line 104: can we make this just one image? with different entrypoints?

charts/langsmith/values.yaml
```diff
@@ -1158,6 +1182,391 @@ aceBackend:
...
+  # -- Deploy the in-chart SmithDB workloads.
+  enabled: false
+  # -- Name segment used for SmithDB resources.
+  name: "smithdb"
+  commonLabels: {}
+  commonAnnotations: {}
+
+  config:
+    # -- Environment name used for SmithDB object-store defaults and labels.
+    env: "self-hosted"
```
langchain-infra: why add this?

charts/langsmith/values.yaml
```diff
@@ -1158,6 +1182,391 @@ aceBackend:
...
 
+smithdb:
+  # -- Deploy the in-chart SmithDB workloads.
+  enabled: false
+  # -- Name segment used for SmithDB resources.
+  name: "smithdb"
+  commonLabels: {}
+  commonAnnotations: {}
+
+  config:
```
langchain-infra, line 1225: for some of this stuff, should it live in our top level config? i wouldnt expect to configure observability twice for example. i would also expect stuff like commonLabels to propagate across the chart instead of having a separate smithdb commonLabels?

charts/langsmith/templates/smithdb/client-deployment.yaml
ben11211: It's very useful for early access testing.

charts/langsmith/values.yaml
ben11211: Would opt to keep separate for now because SmithDB only supports gRPC export.

charts/langsmith/values.yaml
```diff
@@ -1158,6 +1182,379 @@ aceBackend:
...
+        capabilities:
+          drop:
+            - ALL
+      resources: {}
+      # -- Optional probe override. Empty uses chart defaults.
+      probes: {}
+      command:
+        - "./smithdb-query"
+      extraEnv:
+        - name: SMITHDB_QUERY__GRPC_PORT
```
langchain-infra: why set this here? can we inline it?

charts/langsmith/values.yaml
langchain-infra: i dont think i follow? what does grpc export mean here?

charts/langsmith/values.yaml
```diff
@@ -1158,6 +1182,379 @@ aceBackend:
...
+        accessKeyIdSecretKey: ""
+        secretAccessKeySecretKey: ""
+
+  langsmith:
+    dualIngest:
+      # -- Enables LangSmith run ingestion to also write to SmithDB. Requires smithdb.enabled.
+      enabled: false
+    frontend:
+      # -- Defaults the LangSmith UI to SmithDB endpoints for run queries. Requires smithdb.enabled.
+      useSmithDBEndpoints: false
```
ben11211, line 1266: Considering if these should be moved to top level.

charts/langsmith/values.yaml
ben11211: This is being tracked, and will be handled prior to general availability.

charts/langsmith/templates/smithdb/client-deployment.yaml
langchain-infra: can we deploy it separately? outside our chart?

charts/langsmith/templates/smithdb/compaction-worker-deployment.yaml
```diff
@@ -0,0 +1,110 @@
+{{- if and .Values.smithdb.enabled .Values.smithdb.compactionWorker.enabled }}
+{{- $workerEnv := list
+  (dict "name" "SMITHDB_COMPACTION_WORKER__HTTP_PORT" "value" (.Values.smithdb.compactionWorker.containerPort | toString))
+  (dict "name" "SMITHDB_COMPACTION_WORKER__GRPC_ENDPOINT" "value" (include "langsmith.smithdb.httpServiceUrl" (dict "root" . "component" "compaction" "port" .Values.smithdb.compaction.service.port)))
+  (dict "name" "SMITHDB_COMPACTION_WORKER__GRPC_REQUEST_TIMEOUT" "value" "30s")
+-}}
+{{- with .Values.smithdb.compactionWorker.maxConcurrentJobs }}
+{{- $workerEnv = append $workerEnv (dict "name" "SMITHDB_COMPACTION_WORKER__MAX_CONCURRENT_JOBS" "value" (. | toString)) -}}
```
langchain-infra: why not just inline these env vars? this is hard to read / understand

charts/langsmith/templates/_helpers.tpl
langchain-infra: this doesnt seem resolved?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -487,6 +487,313 @@ Template containing common environment variables that are used by several servic
...
+{{- $component := .component -}}
+{{- $port := .port -}}
+{{- printf "%s.%s.svc.%s:%v" (include "langsmith.smithdb.componentName" (dict "root" $root "component" $component)) ($root.Values.namespace | default $root.Release.Namespace) $root.Values.clusterDomain $port }}
+{{- end }}
+
+{{/*
+SmithDB HTTP-style endpoint used by SmithDB services.
+Args: root, component, port.
+*/}}
+{{- define "langsmith.smithdb.httpServiceUrl" -}}
```
langchain-infra, line 536: isnt this in the config-map? why do we still need it?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -487,6 +487,313 @@ Template containing common environment variables that are used by several servic
...
+SmithDB resource name prefix.
+*/}}
+{{- define "langsmith.smithdb.fullname" -}}
+{{- printf "%s-%s" (include "langsmith.fullname" .) .Values.smithdb.name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common SmithDB labels.
+*/}}
+{{- define "langsmith.smithdb.labels" -}}
```
langchain-infra: why have this? just use langsmith.labels?

charts/langsmith/templates/_helpers.tpl
langchain-infra: oh i see the override logic, i think i was just suggesting that we add the probes in the values.yaml file and pipe in directly. user can edit anything in that values.yaml file

charts/langsmith/values.yaml
langchain-infra: imo yes

charts/langsmith/values.yaml
```diff
@@ -1186,6 +1210,382 @@ aceBackend:
...
+    clusterManager:
+      # -- Lets LangSmith services call the SmithDB cluster manager. Requires smithdb.clusterManager.enabled.
+      enabled: false
```
langchain-infra: when would we turn off clusterManager?

charts/langsmith/values.yaml
```diff
@@ -1186,6 +1210,382 @@ aceBackend:
...
+        fsGroup: 1001
+      securityContext:
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+            - ALL
+      resources: {}
+      # -- Optional probe override. Empty uses chart defaults.
+      probes: {}
+      command:
```
langchain-infra, line 1306: longer term, we should move to entrypoint.sh scripts as it allows people to wrap more easily if they need to add other logic

charts/langsmith/values.yaml
```diff
@@ -1186,6 +1210,382 @@ aceBackend:
...
+    deployment:
+      replicas: 1
+      strategy:
+        type: RollingUpdate
+        rollingUpdate:
+          maxSurge: 1
+          maxUnavailable: 0
+      labels: {}
+      annotations: {}
+      podSecurityContext:
```
langchain-infra: any reason to set by default? we typically just runAsRoot by default and then ahve the read-only example for people who want to run as unprivileged/as a specific user

langchain-infra (commented):
some comments, mostly nits. i think generally there is just a lot of logic in helm, i like generally having dumb templates when we can (that just read from values.yaml directly essentially)

charts/langsmith/templates/_helpers.tpl
ben11211: This provides address consistency despite differing env keys for smithdb components and langsmith components. 

For SmithDB<->SmithDB in-cluster connectivity, it's this style
`SMITHDB_<COMPONENT>__GRPC_ENDPOINT`
`SMITHDB_<COMPONENT>__CLUSTER_MANAGER__ENDPOINT`

for Langsmith its 
`SMITHDB_CLUSTER_MANAGER_URL` 
`SMITHDB_QUERY_SERVICE_URL`
`SMITHDB_MUTATIONS_SERVICE_URL` 

I think you're pointing towards mounting the general config-map onto SmithDB components and duplicating these values within the config-map. IMO isolating SmithDB component-specific configuration is a good idea considering it's generally component specific. I'd opt to avoid supporting sharing env vars between SmithDB and Langsmith if possible. For cases where we need consistent behavior across that boundary, I'd prefer to use chart config for flexibility. We can change this later if we decide to rework our config system prior to end of EA?

charts/langsmith/values.yaml
ben11211: Will look to do this prior to release. I think premature now.

charts/langsmith/values.yaml
ben11211: SmithDB doesn't support exporting over OTLP/HTTP (only OTLP/gRPC), which is the langsmith chart default. We can consolidate once we add OTLP/HTTP support to SmithDB, but right now adopting that would result in bad defaults for SmithDB tracing.

However - pulled in the labels/annotations.

charts/langsmith/templates/_helpers.tpl
ben11211: Addressed here https://github.com/langchain-ai/helm/pull/736/changes#r3397791090

langchain-infra (approved):
approved just pending my one nit

PR #739, opened 2026-05-28
chart(langsmith): release 0.15.1 — drop insights/polly bootstrap, enable by default

langchain-infra:
## Summary

Removes the bundled-agent bootstrap path for Insights and Polly. Both now run as standard Helm-managed Deployments via the top-level \`insights\` / \`polly\` value blocks (same shape as \`fleet\`).

- \`insights.enabled\` and \`polly.enabled\` default to \`true\` — fresh installs get the standalone stacks (api-server, queue, bundled Postgres + Redis) out of the box.
- \`config.insights.enabled\` and \`config.polly.enabled\` are no longer supported; templating hard-fails with a message pointing at the top-level blocks.

Chart version: \`0.15.0\` → \`0.15.1\`. appVersion stays at \`0.15.8\`.

## Heads-up

Fresh installs without an \`existingSecretName\` now need to supply \`insights.encryptionKey\` and \`polly.encryptionKey\` (or set the respective \`.enabled: false\`).

## Test plan

- [ ] \`helm unittest charts/langsmith\` passes
- [ ] \`helm lint charts/langsmith\` passes
- [ ] Default \`helm template\` renders both standalone insights and polly stacks
- [ ] \`helm template --set insights.enabled=false --set polly.enabled=false\` cleanly skips both
- [ ] \`helm template --set config.insights.enabled=true\` fails with the deprecation error
- [ ] \`helm upgrade\` against a release that previously used \`config.insights.enabled=true\` / \`config.polly.enabled=true\` tears down the bundled-agent Deployments and brings up the new standalone ones

### DIFF
--- a/charts/langsmith/values.yaml
+++ b/charts/langsmith/values.yaml
@@ -849,37 +849,6 @@ config:
     # Disable this if there is restricted access to the external facing ingress endpoint for your agent(s) from inside of your kubernetes cluster.
     ingressHealthCheckEnabled: true
 
-  # Insights (Clio) configuration - AI-powered insights agent
-  # Requires config.deployment.enabled to be true
-  insights:
-    enabled: false
-    # Fernet encryption key for Insights. Required when enabled.
-    # Generate with: python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
-    encryptionKey: ""
-    agent:
-      resources:
-        cpu: 2
-        cpuLimit: 4
-        memoryMb: 4096
-        memoryLimitMb: 8192
-        minScale: 1
-        maxScale: 5
-      extraEnv: {}
-
-  # Polly configuration
-  polly:
-    enabled: false
-    encryptionKey: ""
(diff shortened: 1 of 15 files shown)


charts/langsmith/values.yaml
```diff
@@ -849,8 +849,6 @@ config:
     # Disable this if there is restricted access to the external facing ingress endpoint for your agent(s) from inside of your kubernetes cluster.
     ingressHealthCheckEnabled: true
 
-  # DEPRECATED: use the top-level insights block instead. This bundled-agent
-  # path is managed by backend.agentBootstrap and is being phased out.
```
joaquin-borggio-lc, line 853: should we remove the blocks from the values file entirely? or in follow up?

joaquin-borggio-lc (approved):
one comment

charts/langsmith/values.yaml
langchain-infra: done

PR #757, opened 2026-06-08
feat: redact sensitive data in diagnostics

gethin-langchain:
**Summary**
Updates get_k8s_debugging_info.sh to expand default diagnostic collection and adds flags to control what gets captured and whether the bundle is scrubbed before sharing.

- kubectl describe for every pod
- kubectl logs per container — last 24h plus previous logs on restart
- helm get values per release in the namespace (auto-discovered, skipped if helm not installed)
- Existing: kubectl get all, kubectl get events, kubectl top pods

**New flags**
--redact — opt-in best-effort scrub of the entire bundle before compression
--exclude-logs — skip pod log collection
--exclude-describe — skip kubectl describe collection

**What --redact scrubs**
- Secrets: API keys (OpenAI, Anthropic, AWS, GCP, GitHub, Stripe, Slack, Twilio, LangSmith), JWTs, Bearer/Authorization headers, URI passwords, PEM private keys
- Identifiers: emails, phones, SSNs, card numbers, UUIDs, IPs, MACs, dates, ICD-10 codes, tenant/user/patient ID headers
- Trace data: inputs, outputs, messages, content, tool_calls, metadata, etc.

Notes
- Diagnostic data (errors, memory/queue stats, pod lifecycle, K8s events) is preserved
- --redact applies to every collected file — logs, describe output, resource YAML, Helm values
- Best-effort only — won't catch free-text PHI (names, addresses) in logs, and can mangle timestamps/IPs/hashes. A stderr banner warns users to review the bundle before sharing
- A warning is also printed when running without --redact to remind users the bundle may contain raw customer data

### DIFF
--- a/charts/langsmith/scripts/get_k8s_debugging_info.sh
+++ b/charts/langsmith/scripts/get_k8s_debugging_info.sh
@@ -1,23 +1,169 @@
 #!/bin/bash
 
 # We expect the namespace hosting all kubernetes resources to be passed as an argument to this script
+REDACT=0
+
 while [[ "$#" -gt 0 ]]; do
   case $1 in
     --namespace) NS="$2"; shift ;;
+    --redact) REDACT=1 ;;
     *) echo "Unknown parameter passed: $1"; exit 1 ;;
   esac
   shift
 done
 
 if [ -z "$NS" ]; then
-  echo "Usage: $0 --namespace <namespace>"
+  echo "Usage: $0 --namespace <namespace> [--redact]"
   exit 1
 fi
+
+# Best-effort redaction patterns. Masks structured secrets and PII only.
+# Does NOT remove free-text PHI (names, addresses, DOBs, diagnoses) that may
+# appear inside application log messages — review the bundle before sharing.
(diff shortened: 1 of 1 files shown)


charts/langsmith/scripts/get_k8s_debugging_info.sh
```diff
@@ -1,19 +1,200 @@
...
+      print
+    }
+  ' "$f" > "$tmp" && mv "$tmp" "$f"
+}
+
+# ---------------------------------------------------------------------------
+# Warnings
+# ---------------------------------------------------------------------------
+
+if [[ "$REDACT" -eq 0 ]]; then
```
l2and, line 164: Since this is higher risk, I think it's worthwhile making the user confirm that `REDACT=0` is really what they want with like a "y to continue".

l2and (commented):
Just adding in some additional checks to make sure a user wants to proceed.

ddzmitry (approved):

langchain-infra (approved):

PR #761, opened 2026-06-09
feat(langsmith): expose remote MCP server + OAuth AS under /api

langchain-infra:
## Summary

Exposes the OAuth-protected **remote MCP server** (and its OAuth 2.1 Authorization Server) under the self-hosted `/api` route prefix, so external MCP clients (Claude, Cursor, …) can connect to a self-hosted/BYOC LangSmith. Today this subsystem is only reachable on SaaS (everything at host root); self-hosted routes everything under `/api(/v1)` and never proxied/advertised the OAuth+MCP+well-known endpoints with a path-prefixed issuer.

## Enablement model

- **Automatic on `config.hostname`.** When a hostname is set, the chart sets `OAUTH_AS_ENABLED=true`, derives `OAUTH_AS_ISSUER` (`<hostname>[/basePath]/api`), and adds the `/api` nginx routes.
- **Activates only when a signing JWKS is present.** smith-go self-gates (`OAuthASEnabled && signer != nil`, `main.go:751`), so without the signing key the endpoints are **inert** (they 404). No client can use them without a key.
- **The chart owns `LANGSMITH_SIGNING_JWKS`** (modeled on `api_key_salt`): wired by default from the signing JWKS secret — `config.signingJwks` (inline → chart-managed secret) or key `langsmith_signing_jwks` in `config.existingSecretName`. Operators **must not** set it manually via `commonEnv`/`extraEnv`; `validate.yaml` fails with a clear message (a skip-guard keeps the generic duplicate-key error from masking it).

## New values

- `config.signingJwks` — optional inline Ed25519/OKP JWKS (RSA rejected).
- `config.oauthAsIssuer` — optional issuer override (defaults to `<hostname>[/basePath]/api`).
- `config.llmAuthProxyIssuer` — optional `LLM_AUTH_PROXY_ISSUER` (emitted only when set).

## nginx routes (both basePath and non-basePath blocks)

`/api/mcp` (MCP JSON-RPC, distinct from the root `/mcp` fleet tool server) · `/api/oauth/` · `/api/.well-known/oauth-protected-resource/mcp` (RFC 9728) · authorization-server metadata in RFC 8414 path-insertion (`/.well-known/oauth-authorization-server/api`) **and** path-append forms · `/api/.well-known/jwks.json`.

## ⚠️ Behavior changes on the v15 train

1. **OAuth AS + `/api` routes now render for every install with a `config.hostname`** (≈all). They're **inert without a signing JWKS** (smith-go self-gates; routes 404). Installs that *do* have a signing JWKS (e.g. LLM-auth-proxy users) will have remote MCP **enabled** after upgrade — the endpoints are OAuth-protected.
2. **Forced migration:** any install currently setting `LANGSMITH_SIGNING_JWKS` via `commonEnv`/`extraEnv` will **fail the render** on upgrade until they remove it (the chart owns it now; put the key in `config.existingSecretName` / `config.signingJwks`).

No `langchainplus` (smith-go) change required — `OAUTH_AS_ISSUER` is set explicitly.

## Testing

`helm unittest` (29 tests incl. new `tests/remote_mcp_test.yaml`), `helm lint`, and `nginx -t` on the rendered frontend config (feature on) all pass. Chart `0.15.3 → 0.15.4`.

Targets `v15-stable` for the internal dogfood AKS deploy.

### DIFF
--- a/charts/langsmith/Chart.yaml
+++ b/charts/langsmith/Chart.yaml
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
-version: 0.15.3
+version: 0.15.4
 appVersion: "0.15.9"
--- a/charts/langsmith/templates/_helpers.tpl
+++ b/charts/langsmith/templates/_helpers.tpl
@@ -974,6 +974,18 @@ which default to http:// for local development.
 {{- end -}}
 
 {{/*
+OAuth issuer URL advertised by the remote MCP server / OAuth Authorization Server.
+Defaults to <hostnameWithProtocol>[/basePath]/api; overridable via config.remoteMcp.issuer.
+*/}}
+{{- define "langsmith.remoteMcpIssuer" -}}
+{{- if .Values.config.remoteMcp.issuer -}}
+{{- .Values.config.remoteMcp.issuer -}}
+{{- else -}}
+{{- include "langsmith.hostnameWithProtocol" . }}{{- with .Values.config.basePath }}/{{ . }}{{- end }}/api
+{{- end -}}
+{{- end -}}
+
+{{/*
 Public URL for the default Agent Builder MCP server.
 Served through the frontend at /mcp (or /<basePath>/mcp).
 */}}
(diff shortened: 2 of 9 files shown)


langchain-infra:
Superseded by #762, which targets main (ships in v16) instead of v15-stable — avoids forcing the LANGSMITH_SIGNING_JWKS migration onto the v15 customer train. Dogfooding moved to the EKS langsmith-eks-self-hosted-ci cluster (tracks helm@main).

PR #763, opened 2026-06-09
fix(langsmith): route root /oauth/device/authorize for device-flow activation

langchain-infra:
## Summary

Follow-up to [#762](https://github.com/langchain-ai/helm/pull/762). The OAuth **device flow** (used by `langsmith auth login` and other CLI clients) failed at the activation step on self-hosted: the `/activate` page showed **"Invalid or expired code"** even with a valid code.

## Root cause

The device-activation page (SPA at `/activate`) POSTs to the Authorization Server's **root** `/oauth/device/authorize`. In single-origin self-hosted, the SPA's backend origin is empty, so the request goes to the root path — but #762 only routed the two *consent* root paths (`/oauth/authorize/approve`, `/oauth/client/`), not the device one. The POST fell through to the SPA `location /` (405, non-JSON), and the page's fallback error is literally "Invalid or expired code." The request never reached platform-backend.

Verified on the dogfood:

```
POST /oauth/device/authorize        -> 405   (SPA fallback — the bug)
POST /api/oauth/device/authorize    -> 401   (reaches backend, as expected)
POST /oauth/authorize/approve       -> 401   (consent, already routed in #762)
```

## Fix

Route root `location = /oauth/device/authorize` to platform-backend (both basePath and non-basePath branches), mirroring the consent routes. Bumps chart `0.16.0-rc.1 -> 0.16.0-rc.2`.

This is the complete set of root `/oauth/*` browser endpoints — the frontend only calls `/oauth/authorize/approve`, `/oauth/client/`, and `/oauth/device/authorize` at root; the first two were covered, this adds the third.

## Test Plan

- [x] `helm unittest` — added assertions for both root `/oauth/authorize/approve` and `/oauth/device/authorize`
- [ ] Re-run `langsmith auth login --api-url https://<host>/api` against a deployment with this chart and confirm the `/activate` code is accepted

### DIFF
--- a/charts/langsmith/Chart.yaml
+++ b/charts/langsmith/Chart.yaml
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
-version: 0.16.0-rc.1
+version: 0.16.0-rc.2
 appVersion: "0.16.1rc1"
--- a/charts/langsmith/templates/frontend/config-map.yaml
+++ b/charts/langsmith/templates/frontend/config-map.yaml
@@ -371,9 +371,9 @@ data:
             proxy_pass http://{{ include "langsmith.fullname" . }}-{{ .Values.platformBackend.name }}.{{ .Values.namespace | default .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.platformBackend.service.port }};
         }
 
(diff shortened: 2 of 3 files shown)


langchain-infra:
Superseded by #764. Rather than add another root `/oauth/*` route, the SPA now calls the OAuth AS at `/api/oauth/*` (the issuer path, same routes external clients use) via the frontend `oauthBaseUrl` change (langchain-ai/langchainplus#27066), so the device-flow endpoint is reached through the existing `/api/oauth/` route and the root device route isn't needed.

PR #766, opened 2026-06-11
chore: sync v15-stable chart to last v15 release (0.15.7 / 0.15.13)

langchain-infra:
## Summary

helm `v15-stable` was lagging — its chart was at `0.15.3` / appVersion `0.15.9`, while the last v15 release that landed on `main` is chart `0.15.7` / appVersion `0.15.13` (#755, `chore: bump v15-stable to 0.15.7/0.15.13`). `main` has since moved on to the `0.16` line, leaving `v15-stable` pointing at the wrong v15.

This makes `charts/langsmith/` on `v15-stable` exactly match `main` at that last-v15 commit (`adcf8ea13`), so the branch points at the correct v15 release.

- Chart.yaml: `version 0.15.3 → 0.15.7`, `appVersion 0.15.9 → 0.15.13`
- values.yaml: LangSmith image tags → `0.15.13`
- Ports the v15 chart changes that landed on `main` (PDB templates, frontend config-map, read-only config examples, regenerated README, new pdb test)

Scope is limited to `charts/langsmith/`. Per the intent to "mark v15-stable as the last v15", this replaces v15-stable's prior chart content rather than merging it.

## Self-Hosted Release Note

none

## Test Plan

- [ ] CI on this PR (chart lint / template tests) passes
- [ ] Confirm chart `0.15.7` / appVersion `0.15.13` is the intended target before merge


langchain-infra:
Superseded — force-reset v15-stable directly to adcf8ea13 (the last v15 release on main, chart 0.15.7 / appVersion 0.15.13) instead of merging this chart-only PR.

PR #782, opened 2026-06-24
feat(langsmith): allow uncapping CPU/memory limits for self-hosted deployments

saad-supports-langchain:
## Summary

Exposes a chart value to lift the CPU/memory caps on **LangSmith Deployments** (LangGraph Platform) resource specs for self-hosted installs.

- Adds `config.deployment.uncappedResourcesEnabled` (default `false`), wired to the `ENABLE_UNCAPPED_LGP_DEPLOYMENT_RESOURCES` env var through `langsmith.commonEnv` — following the existing `ENABLE_LGP_DEPLOYMENT_HEALTH_CHECK` / `config.deployment.ingressHealthCheckEnabled` pattern (both are host-backend settings).
- When enabled, the host-backend's `ResourceSpec` validators skip the SaaS **upper** caps (e.g. 8/16 cores, 16–32 GB). Floors (0.1 CPU, 128/512 MB) and the 128 MB granularity are still enforced.
- Default `false` matches the backend default, so existing installs are unaffected.

Companion to langchain-ai/langchainplus#27405, which added the `ENABLE_UNCAPPED_LGP_DEPLOYMENT_RESOURCES` setting to `host-backend`.

## New values

`config.deployment.uncappedResourcesEnabled` (bool, default `false`).

## Self-Hosted Release Note

Self-hosted operators can now set `config.deployment.uncappedResourcesEnabled: true` to lift the CPU/memory upper caps on LangSmith Deployments (LangGraph Platform) resource specs, allowing agents larger than the default SaaS limits. Minimum sizes and 128 MB granularity still apply. Defaults to `false`.

## Test Plan

- [x] `helm template` with `--set config.deployment.uncappedResourcesEnabled=true` renders `ENABLE_UNCAPPED_LGP_DEPLOYMENT_RESOURCES` with `value: "true"`; default render yields `value: "false"` (present on all component deployments via `commonEnv`)
- [x] `helm lint` passes
- [ ] `helm unittest` (run in CI; no existing snapshot references this env var)

### DIFF
--- a/charts/langsmith/README.md
+++ b/charts/langsmith/README.md
@@ -1178,9 +1178,10 @@ For information on how to use this chart, up-to-date release notes, and other gu
 | config.customErrorSupportMessage | string | `""` | Custom error support message displayed on error pages (plain text). If empty, defaults to the built-in support messages linking to our Support Portal (https://support.langchain.com). |
 | config.customLogo | object | `{"coBrandingEnabled":true,"enabled":false,"logoUrl":""}` | Custom logo configuration. If enabled, the logoUrl and coBrandingEnabled values must be provided. The logoUrl must be a valid URL to an image like png, jpg, or svg. Co-branding will show LangSmith and customer logos side by side. |
 | config.defaultWorkspaceName | string | `"Workspace 1"` | Default workspace name to be provisioned when org is created. |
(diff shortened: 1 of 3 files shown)


hari-dhanushkodi (approved):
nit, product name is technically `LangSmith Deployment` instead of plural form

saad-supports-langchain:
> nit, product name is technically `LangSmith Deployment` instead of plural form

fixed

langchain-infra (approved):

PR #784, opened 2026-06-26
Add self-hosted sandbox runtime support

DanielKneipp:
## Summary

- Add optional LangSmith Sandboxes support behind `config.sandboxes.enabled` for the same-cluster `sandbox-host` architecture.
- Deploy sandbox runtime resources in the LangSmith release namespace, including the `sandbox-host` Deployment, service account/RBAC, proxy CA resources, JuiceFS CSI dependency, and sandbox JuiceFS PV/PVCs.
- Wire LangSmith services to sandboxes when enabled, including platform/ingest env vars, frontend sandbox routes, service auth and callback signing through the standard LangSmith app Secret, and optional sandbox service URLs via `config.sandboxes.serviceUrlBaseUrl`.
- Configure sandbox storage through JuiceFS CSI with explicit JuiceFS Redis metadata config, supported storage validation for AWS/EKS and GCP/GKE, flat JuiceFS volume-name validation, and pre-bound static PV/PVCs.
- Add sandbox chart validation and unit coverage for required image tags, cluster name, secrets, env conflicts, scheduling, proxy CA modes, JuiceFS config, PV/PVC binding, and existing Secret flows.
- Update chart packaging/release workflows so the JuiceFS CSI dependency is available when linting, packaging, and publishing the chart.

## Support Notes

- Self-hosted sandboxes currently support AWS/EKS and GCP/GKE.
- Azure/AKS remains supported by the base LangSmith chart, but not by sandboxes yet.
- Sandboxes are disabled by default.
- Private IP sandbox callbacks default to disabled and can be explicitly enabled with `config.sandboxes.sandboxHost.allowPrivateIPCallbacks`.
- The chart installs the JuiceFS CSI dependency when sandboxes are enabled. Only one sandbox-enabled LangSmith release should manage the JuiceFS CSI cluster-scoped resources per Kubernetes cluster.
- `sandbox-host` runs as a Deployment with a no-surge rolling update strategy so host pods roll one at a time on host-networked nodes.

## Validation

- `helm unittest charts/langsmith`
- `helm dependency build charts/langsmith`
- `helm template test charts/langsmith -f charts/langsmith/tests/values/sandboxes-enabled.yaml`
- `helm lint charts/langsmith`
- `kubeconform -summary -ignore-missing-schemas /tmp/langsmith-sandboxes-render.yaml`
- `git diff --check`


charts/langsmith/templates/sandboxes/juicefs-csi.yaml
```diff
@@ -0,0 +1,111 @@
...
+  storageClassName: ""
+  csi:
+    driver: csi.juicefs.com
+    volumeHandle: {{ .Values.config.sandboxes.juicefs.csi.pvName }}
+    fsType: juicefs
+    volumeAttributes:
+      juicefs/mount-service-account: juicefs-csi-node-sa
+    nodePublishSecretRef:
+      name: {{ include "langsmith.sandboxes.juicefsCSIConfigSecretName" . }}
+      namespace: {{ $namespace }}
```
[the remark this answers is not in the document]
DanielKneipp: Valid finding. Fixed in 265386c by adding claimRef to both static JuiceFS PVs and volumeName to the matching PVCs, with tests covering the regular and host PV/PVC bindings.

charts/langsmith/templates/frontend/config-map.yaml
```diff
@@ -146,6 +146,27 @@ data:
...
+            rewrite /{{ .Values.config.basePath }}/api/v2/sandboxes/(.*) /v2/sandboxes/$1  break;
+            proxy_set_header Upgrade $http_upgrade;
+            proxy_set_header Connection $connection_upgrade;
+            proxy_http_version 1.1;
+            proxy_buffering off;
+            proxy_cache off;
+            proxy_pass http://{{ include "langsmith.fullname" . }}-{{ .Values.platformBackend.name }}.{{ .Values.namespace | default .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.platformBackend.service.port }};
+        }
+        {{- end }}
+
```
langchain-infra, line 151: why is this needed?

charts/langsmith/templates/frontend/config-map.yaml
```diff
@@ -780,6 +801,27 @@ data:
...
+        location /api/v2/runs/analytics {
```
langchain-infra: same?

charts/langsmith/templates/frontend/deployment.yaml
```diff
@@ -51,7 +51,7 @@
 - name: VITE_CUSTOM_ERROR_SUPPORT_MESSAGE
   value: {{ .Values.config.customErrorSupportMessage | quote }}
 {{- end }}
-{{- if .Values.config.deployment.enabled }}
+{{- if or .Values.config.deployment.enabled .Values.config.sandboxes.enabled }}
```
langchain-infra: does sandboxes need vite host? we dont use host backend for this anymore right?

charts/langsmith/templates/ingest-queue/deployment.yaml
```diff
@@ -5,7 +5,7 @@
   value: {{ .Values.ingestQueue.containerPort | quote }}
 {{- end -}}
 {{- if .Values.ingestQueue.enabled }}
-{{- $envVars := concat .Values.commonEnv (include "langsmith.commonEnv" . | fromYamlArray) (include "ingestQueueEnvVars" . | fromYamlArray) .Values.ingestQueue.deployment.extraEnv -}}
+{{- $envVars := concat .Values.commonEnv (include "langsmith.commonEnv" . | fromYamlArray) (include "ingestQueueEnvVars" . | fromYamlArray) (include "langsmith.sandboxes.ingestQueueEnv" . | fromYamlArray) .Values.ingestQueue.deployment.extraEnv -}}
```
langchain-infra: nit, should we just update the ingestQueueEnvVars function? instead of a separate ingestQueueEnv param?

charts/langsmith/templates/platform-backend/deployment.yaml
```diff
@@ -14,7 +14,7 @@
       key: langsmith_signing_jwks
       optional: true
 {{- end -}}
-{{- $envVars := concat .Values.commonEnv (include "langsmith.commonEnv" . | fromYamlArray) (include "platformBackendEnvVars" . | fromYamlArray) .Values.platformBackend.deployment.extraEnv -}}
+{{- $envVars := concat .Values.commonEnv (include "langsmith.commonEnv" . | fromYamlArray) (include "platformBackendEnvVars" . | fromYamlArray) (include "langsmith.sandboxes.platformBackendEnv" . | fromYamlArray) .Values.platformBackend.deployment.extraEnv -}}
```
langchain-infra: same here

charts/langsmith/templates/sandboxes/sandbox-host-installer-daemonset.yaml
```diff
@@ -0,0 +1,93 @@
...
+            - -binary-path
+            - /host-bin/sandbox-host
+            - -ready-file
+            - /tmp/installer-ready
+            - -timeout
+            - {{ .Values.config.sandboxes.sandboxHostInstaller.daemonSet.timeout | quote }}
+          env:
+            - name: LANGCHAIN_ENV
+              value: {{ .Values.config.langchainEnv | quote }}
+            - name: SANDBOX_X_SERVICE_AUTH_JWT_SECRET
```
langchain-infra: in the new arch, do we need separate jwt for sandboxes?

charts/langsmith/templates/sandboxes/sandbox-host-rbac.yaml
```diff
@@ -0,0 +1,49 @@
...
+    {{- end }}
+  annotations:
+    {{- include "langsmith.annotations" . | nindent 4 }}
+    {{- with .Values.config.sandboxes.sandboxHost.serviceAccount.annotations }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+automountServiceAccountToken: {{ .Values.config.sandboxes.sandboxHost.serviceAccount.automountServiceAccountToken }}
+---
+{{- end }}
+apiVersion: rbac.authorization.k8s.io/v1
```
langchain-infra, line 3: some people will probably want to be able to pass these values in for rbac, i think you can copy how we do rbac for the operator

charts/langsmith/templates/sandboxes/sandbox-host-statefulset.yaml
```diff
@@ -0,0 +1,151 @@
...
+        {{- include "langsmith.sandboxes.sandboxHostNodeSelector" . | nindent 8 }}
+      {{- with .Values.config.sandboxes.sandboxHost.statefulSet.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: sandbox-host
+          image: {{ include "langsmith.image" (dict "Values" .Values "Chart" .Chart "component" "sandboxHostImage") | quote }}
+          imagePullPolicy: {{ .Values.images.sandboxHostImage.pullPolicy }}
+          command:
```
langchain-infra: i think we should try and move all commands into entrypoint scripts so people can potentially wrap. (and make them able to be passed in)

charts/langsmith/templates/sandboxes/sandbox-host-deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+            - -c
+            - |
+              set -e
+              if [ ! -x /host-bin/sandbox-host ]; then
+                cp /opt/sandbox-host/bin/sandbox-host /host-bin/sandbox-host.tmp
+                mv /host-bin/sandbox-host.tmp /host-bin/sandbox-host
+              fi
+              exec /host-bin/sandbox-host
+          env:
+            - name: LANGCHAIN_ENV
```
langchain-infra, line 88: is this a value we expose? i believe typically we just always set .env.local_k8s

charts/langsmith/templates/sandboxes/sandbox-host-statefulset.yaml
```diff
@@ -0,0 +1,151 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "langsmith.secretsName" . }}
+                  key: sandbox_callback_signing_jwk
+                  optional: true
+            {{- end }}
+            - name: OAUTH_AS_ISSUER
+              value: {{ include "langsmith.sandboxes.langsmithInternalEndpoint" . | quote }}
+            - name: SSRF_ALLOW_PRIVATE_IPS_CALLBACKS
+              value: "true"
```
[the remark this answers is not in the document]
langchain-infra: lets set this in .env

charts/langsmith/templates/sandboxes/smithbox-control.yaml
```diff
@@ -0,0 +1,228 @@
+{{- if .Values.config.sandboxes.enabled }}
```
langchain-infra: in final state, do we still need control? i think control is unused?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1185,6 +1185,191 @@ Strip protocol (http://, https://, etc.) from hostname
...
+
+{{/*
+smithbox-control in-cluster URL.
+*/}}
+{{- define "langsmith.sandboxes.smithboxControlUrl" -}}
+{{- printf "http://%s.%s.svc.%s:%v" .Values.config.sandboxes.smithboxControl.name (.Values.namespace | default .Release.Namespace) .Values.clusterDomain .Values.config.sandboxes.smithboxControl.service.port -}}
+{{- end -}}
+
+{{/*
+Firecracker size classes rendered in the JSON shape expected by smithbox-control.
```
langchain-infra: dont think this is needed anymore

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1185,6 +1185,191 @@ Strip protocol (http://, https://, etc.) from hostname
...
+{{- toYaml .Values.config.sandboxes.sandboxHost.statefulSet.nodeSelector -}}
+{{- else -}}
+sandbox.langsmith.com/host: "true"
+{{- end -}}
+{{- end -}}
+
+{{/*
+LangSmith app env vars for sandbox support.
+*/}}
+{{- define "langsmith.sandboxes.platformBackendEnv" -}}
```
langchain-infra: can we just add these to the existing env vars? but toggle with the var

charts/langsmith/templates/config-map.yaml
```diff
@@ -33,8 +33,11 @@ data:
   VITE_BACKEND_AUTH_TYPE: "none"
   AUTH_TYPE: "none"
   {{- end }}
-  LANGCHAIN_ENV: "local_kubernetes"
+  LANGCHAIN_ENV: {{ .Values.config.langchainEnv | quote }}
```
langchain-infra: why make this configurable?

charts/langsmith/Chart.yaml
```diff
@@ -7,3 +7,8 @@ description: Helm chart to deploy the langsmith application and all services it
 type: application
 version: 0.16.0-rc.8
 appVersion: "0.16.10rc1"
+dependencies:
```
langchain-infra: i see us create this manually, what does the subchart also install?

charts/langsmith/values.yaml
```diff
@@ -915,6 +927,159 @@ config:
...
+    # When config.existingSecretName is set, add key `sandbox_x_service_auth_jwt_secret` to that Secret instead.
+    xServiceAuthJwtSecret: ""
+    # -- Optional previous service-auth secret used during rotation. Used only when this chart creates the LangSmith app Secret.
+    # When config.existingSecretName is set, add optional key `sandbox_x_service_auth_jwt_secret_previous` to that Secret to enable rotation.
+    xServiceAuthJwtSecretPrevious: ""
+    # -- Required private JWK for sandbox callback signing when this chart creates the LangSmith app Secret.
+    # When config.existingSecretName is set, add key `sandbox_callback_signing_jwk` to that Secret instead.
+    callbackSigningJwk: ""
+    # -- Optional default sandbox blueprint/rootfs image.
+    defaultBlueprintImage: ""
```
langchain-infra: think bundled into host

langchain-infra (changes requested):

charts/langsmith/templates/frontend/config-map.yaml
DanielKneipp: needed because the frontend calls /api/v2/sandboxes/... while platform-backend serves /v2/sandboxes/...

charts/langsmith/templates/frontend/deployment.yaml
DanielKneipp: Current FE routes sandboxes under the Host route shell, so VITE_HOST_ENABLED is still needed even though host-backend is not. But I can look at changing that as well if you think is better

charts/langsmith/templates/sandboxes/sandbox-host-installer-daemonset.yaml
DanielKneipp: yep, host still uses it for comms with the api. Host installer also uses it to talk with host itself

charts/langsmith/templates/sandboxes/smithbox-control.yaml
DanielKneipp: will test now internally, but we should not

charts/langsmith/Chart.yaml
DanielKneipp: The subchart installs JuiceFS CSI driver resources. We create the sandbox-specific Secret/PV/PVC config.

langchain-infra: how many resources? can we just inline it?

charts/langsmith/templates/frontend/deployment.yaml
langchain-infra: yea can we just change that in frontend?

charts/langsmith/templates/frontend/config-map.yaml
langchain-infra: sorry i meant the runs/analytics part? or could this not be /api/v2 -> /v2 in platform-backend?

charts/langsmith/templates/platform-backend/deployment.yaml
langchain-infra: i dont think this was changed?

charts/langsmith/templates/ingest-queue/deployment.yaml
langchain-infra: i dont think this was changed/fixed?

charts/langsmith/templates/sandboxes/proxy-ca.yaml
```diff
@@ -0,0 +1,59 @@
+{{- if and .Values.config.sandboxes.enabled (eq .Values.config.sandboxes.proxyCa.mode "generatedSecret") }}
```
langchain-infra: imo, we should just support secret. My understanding is cert manager will also create a secret here? Then people can use this like we do for the other secrets:

1. inline the values, we create the cert
2. point to existingSecretName (which can be generated by a cert manager cert they create or whatever cert provisioning thing they have).

charts/langsmith/templates/sandboxes/sandbox-host-installer-daemonset.yaml
langchain-infra, line 1: think this can be removed also?

charts/langsmith/templates/sandboxes/sandbox-host-installer-daemonset.yaml
```diff
@@ -0,0 +1,93 @@
...
+        {{- include "langsmith.sandboxes.sandboxHostNodeSelector" . | nindent 8 }}
+      {{- with .Values.config.sandboxes.sandboxHost.statefulSet.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: installer
+          image: {{ include "langsmith.image" (dict "Values" .Values "Chart" .Chart "component" "sandboxHostImage") | quote }}
+          imagePullPolicy: {{ .Values.images.sandboxHostImage.pullPolicy }}
+          command:
```
langchain-infra: nit, would be nice to wrap some of these in shell scripts on the image

charts/langsmith/templates/sandboxes/sandbox-host-service.yaml
langchain-infra, line 1: think removable?

charts/langsmith/templates/sandboxes/sandbox-host-deployment.yaml
langchain-infra, line 1: believe this needs to move to deployment now?

charts/langsmith/templates/sandboxes/sandbox-host-deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+            - name: LANGSMITH_SANDBOX_CALLBACK_SIGNING_JWK
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "langsmith.secretsName" . }}
+                  key: sandbox_callback_signing_jwk
+                  optional: true
+            {{- end }}
+            - name: OAUTH_AS_ISSUER
+              value: {{ include "langsmith.sandboxes.langsmithInternalEndpoint" . | quote }}
+            - name: SSRF_ALLOW_PRIVATE_IPS_CALLBACKS
```
langchain-infra, line 132: can we set in .env.kubernertes?

charts/langsmith/templates/sandboxes/smithbox-control.yaml
langchain-infra, line 1: think this is removable?

charts/langsmith/templates/sandboxes/sandbox-host-service.yaml
langchain-infra: ignroe this review, for some reason my gh didnt update?

charts/langsmith/Chart.yaml
DanielKneipp: 16 JuiceFS driver resources from 13 template files. A bit tough to inline.

charts/langsmith/templates/sandboxes/sandbox-host-deployment.yaml
```diff
@@ -0,0 +1,140 @@
...
+        - name: juicefs
+          persistentVolumeClaim:
+            claimName: {{ include "langsmith.sandboxes.juicefsHostPVCName" . }}
+        - name: proxy-ca
+          secret:
+            secretName: {{ include "langsmith.sandboxes.proxyCaSecretName" . }}
+        - name: host-bin
+          hostPath:
+            path: /var/lib/sandbox-host
+            type: DirectoryOrCreate
```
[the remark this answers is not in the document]
DanielKneipp: sandbox-host endpoint requires auth. It is expected of it to have privileged access to the underlying node

.github/workflows/helm_checks.yaml
```diff
@@ -38,6 +38,7 @@ jobs:
       - name: Add Helm Repositories
         run: |
           helm repo add langchain https://langchain-ai.github.io/helm/
+          helm repo add juicedata https://juicedata.github.io/charts/
```
langchain-infra: shouldnt we not need this anymore?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1210,6 +1210,197 @@ Falls back to config.hostname when config.frontendHostname is unset.
...
+*/}}
+{{- define "langsmith.sandboxes.sandboxHostServiceAccountName" -}}
+{{- if .Values.config.sandboxes.sandboxHost.serviceAccount.create -}}
+{{- default (include "langsmith.sandboxes.sandboxHostDeploymentName" .) .Values.config.sandboxes.sandboxHost.serviceAccount.name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- default "default" .Values.config.sandboxes.sandboxHost.serviceAccount.name -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
```
langchain-infra: can we just make this the default in values.yaml? but always set from values.yaml?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1210,6 +1210,197 @@ Falls back to config.hostname when config.frontendHostname is unset.
...
+{{ $imageConfig.repository }}:{{ $imageConfig.tag }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Sandbox-host Deployment name. The host is told this via SANDBOX_HOST_DEPLOYMENT_NAME,
+since it resolves its own Deployment by name to drive pool autoscaling.
+*/}}
+{{- define "langsmith.sandboxes.sandboxHostDeploymentName" -}}
```
langchain-infra: why do we need this? do we do this elsewhere to determine the component name?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1210,6 +1210,197 @@ Falls back to config.hostname when config.frontendHostname is unset.
...
+{{- end -}}
+
+{{- define "langsmith.sandboxes.juicefsCSIDriverConfigChecksum" -}}
+{{- include "langsmith.sandboxes.juicefsCSIDriverConfig" . | sha256sum -}}
+{{- end -}}
+
+{{/*
+Creates an optional image reference without falling back to Chart.AppVersion.
+*/}}
+{{- define "langsmith.optionalImage" -}}
```
langchain-infra: why do we need this?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1210,6 +1210,197 @@ Falls back to config.hostname when config.frontendHostname is unset.
...
+{{- define "langsmith.sandboxes.proxyCaSecretName" -}}
+{{- if eq .Values.config.sandboxes.proxyCa.mode "existingSecret" -}}
+{{- .Values.config.sandboxes.proxyCa.existingSecretName -}}
+{{- else -}}
+{{- default "smithbox-proxy-ca" .Values.config.sandboxes.proxyCa.secretName -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Internal LangSmith platform endpoint used by sandbox runtime callbacks.
```
langchain-infra: why do sandboxes hit both the external and internal apis here?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1228,6 +1419,26 @@ which default to http:// for local development.
 {{- end -}}
 {{- end -}}
 
+{{/*
+Public API endpoint used by sandbox runtime-v2 dataplane URLs.
+When hostname is unset, keep this relative so browser callers use the same origin.
+*/}}
+{{- define "langsmith.sandboxes.platformEndpoint" -}}
```
langchain-infra: why do we need this? dont we already have some private url?

charts/langsmith/README.md
```diff
@@ -8,6 +8,10 @@ Helm chart to deploy the langsmith application and all services it depends on.
 
 For information on how to use this chart, up-to-date release notes, and other guides please check out the [documentation.](https://docs.langchain.com/langsmith/kubernetes)
 
+## Frontend Routing
```
langchain-infra: remove this, we only autogenerate notes. and this isnt needed here.

charts/langsmith/values.yaml
```diff
@@ -899,6 +919,140 @@ config:
     # -- Lift the SaaS CPU/memory caps on LangSmith Deployment resource specs. Opt-in for self-hosted.
     uncappedResourcesEnabled: false
 
+  # -- LangSmith Sandboxes configuration. Disabled by default. When enabled, this installs sandbox runtime resources in the same namespace as the LangSmith release and wires the LangSmith API/frontend services to expose sandbox functionality.
```
langchain-infra: we have way too many comments here. can we lower the amount of information?

charts/langsmith/values.yaml
```diff
@@ -899,6 +919,140 @@ config:
...
+      # -- generatedSecret creates a self-signed CA Secret with Helm and reuses it on live upgrades via lookup. In pure render/GitOps workflows where lookup cannot read the cluster, generatedSecret produces different cert material on each render; use existingSecret for deterministic manifests.
+      mode: "generatedSecret"
+      secretName: "smithbox-proxy-ca"
+      # -- Existing TLS Secret containing tls.crt and tls.key for the sandbox proxy CA. This Secret can be created manually, by cert-manager, or by another external process.
+      existingSecretName: ""
+    juicefs:
+      # -- JuiceFS volume name. Use a flat DNS-label-style name only; slashes and object-store subpaths are not supported here. JuiceFS stores objects under `<name>/` inside the configured bucket. Also used to scope JuiceFS CSI RBAC for the generated mount Secret, so keep it aligned with the `name` key when using an existing CSI config Secret.
+      name: "sandbox-juicefs"
+      # -- Object storage backend used by JuiceFS for sandboxes. Currently supported values are `s3` for AWS/EKS and `gs` for GCP/GKE. Azure-backed sandbox storage is not supported yet.
+      storage: "s3"
```
langchain-infra: can we read this from our blob config?

charts/langsmith/values.yaml
```diff
@@ -899,6 +919,140 @@ config:
...
+      minEphemeralStorageGb: 1
+      maxEphemeralStorageGib: 100
+      maxSandboxes: 1000
+    proxyCa:
+      # -- generatedSecret creates a self-signed CA Secret with Helm and reuses it on live upgrades via lookup. In pure render/GitOps workflows where lookup cannot read the cluster, generatedSecret produces different cert material on each render; use existingSecret for deterministic manifests.
+      mode: "generatedSecret"
+      secretName: "smithbox-proxy-ca"
+      # -- Existing TLS Secret containing tls.crt and tls.key for the sandbox proxy CA. This Secret can be created manually, by cert-manager, or by another external process.
+      existingSecretName: ""
+    juicefs:
```
langchain-infra: can we make these their separate top level sections like we have for each service? add one for host, one for juicefs. also split up

charts/langsmith/values.yaml
```diff
@@ -899,6 +919,140 @@ config:
...
+    # -- Required private JWK for sandbox callback signing when this chart creates the LangSmith app Secret.
+    # When config.existingSecretName is set, add key `sandbox_callback_signing_jwk` to that Secret instead.
+    # Also needs config.hostname for the public issuer; without it, signing fails closed.
+    callbackSigningJwk: ""
+    # -- Optional base URL used to generate browser/programmatic service URLs for HTTP services running inside sandboxes.
+    # When set, the chart renders SANDBOX_SERVICE_URLS for the in-cluster sandbox runtime.
+    # Requires wildcard DNS and TLS for `*.<host>`. When ingress.enabled is true, the chart also adds a wildcard ingress rule to the LangSmith platform backend.
+    # Must be an http:// or https:// URL without a path, query string, or fragment. When ingress.enabled is true, the host must be a DNS hostname without a port.
+    serviceUrlBaseUrl: ""
+    limits:
```
langchain-infra: lets call these quotas to avoid confusion with limits.

charts/langsmith/values.yaml
```diff
@@ -899,6 +919,140 @@ config:
...
+            mountOptions:
+              - cache-dir=/var/cache/juicefs-csi
+              - cache-size=307200
+        pvName: "smithbox-juicefs-csi"
+        pvcName: "smithbox-juicefs-csi"
+    sandboxHost:
+      # -- Component name for sandbox-host. Resources are named `<release-fullname>-<name>`, which the chart passes to the host as SANDBOX_HOST_DEPLOYMENT_NAME.
+      name: "sandbox-host"
+      # -- Whether sandbox callback validation should allow private IP and localhost targets. This renders SSRF_ALLOW_PRIVATE_IPS_CALLBACKS for sandbox-host.
+      allowPrivateIPCallbacks: false
```
langchain-infra: always set this to true, we dont need to make it toggleable

charts/langsmith/templates/platform-backend/deployment.yaml
```diff
@@ -13,6 +13,39 @@
       name: {{ include "langsmith.secretsName" . }}
       key: langsmith_signing_jwks
       optional: true
+{{- if .Values.sandboxes.enabled }}
```
langchain-infra, line 16: add this to the config-map like the other values

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1229,6 +1222,164 @@ Falls back to config.hostname when config.frontendHostname is unset.
...
+{{- .Values.sandboxes.proxyCa.existingSecretName -}}
+{{- else -}}
+{{- default "smithbox-proxy-ca" .Values.sandboxes.proxyCa.secretName -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+In-cluster platform-backend URL. Also rendered as GO_ENDPOINT in the shared ConfigMap.
+*/}}
+{{- define "langsmith.platformBackendEndpoint" -}}
```
langchain-infra, line 1267: why do we need this again?

charts/langsmith/templates/_helpers.tpl
```diff
@@ -1247,6 +1398,26 @@ which default to http:// for local development.
 {{- end -}}
 {{- end -}}
 
+{{/*
+Public API base for sandbox dataplane URLs: scheme, host, any base path, and the /api
+prefix. Relative when no hostname is configured, so browser callers use their own origin.
+*/}}
```
langchain-infra, line 1424: cant we remove this now with the public endpoint?

langchain-infra (approved):

PR #803, opened 2026-07-09
Inline JuiceFS CSI resources for sandboxes

DanielKneipp:
## Summary

- Inline the JuiceFS CSI resources needed by self-hosted sandboxes instead of depending on the upstream `juicefs-csi-driver` chart.
- Add top-level JuiceFS CSI image values so customers can mirror the CSI driver and sidecar images with the rest of LangSmith images.
- Scope inline JuiceFS CSI RBAC so cluster-scoped roles are limited to cluster discovery, while Secret, PVC, job/event, and pod exec/log operations stay scoped to the LangSmith release namespace.
- Run chart-managed JuiceFS CSI without `--cache-client-conf=true`; the bundled setup is intended only for JuiceFS PVs in the namespace where the chart is installed.
- Add checksum annotations so chart-managed JuiceFS config Secret and CSI driver ConfigMap changes roll only the JuiceFS CSI controller/node pods.
- Preserve the existing JuiceFS CSI workload selectors while moving CSI observability and mount-pod configuration to the native sandbox values path.
- Add Helm unit coverage for the inline CSI resources and regenerated chart docs.

Stacked on #784.

## JuiceFS CSI Scope

The bundled JuiceFS CSI setup is intended only for JuiceFS PVs in the namespace where this chart is installed.

This PR intentionally does not pass `--cache-client-conf=true` to the JuiceFS CSI controller. That option enables JuiceFS client-config cache/refresh behavior that is useful for broader CSI installations, but it also requires the driver to discover PV Secret references, read/update JuiceFS config Secrets, and refresh affected mount pods. In practice that pushes the chart toward broader Secret and pod exec RBAC, especially when one CSI installation is expected to serve PVs across namespaces.

For self-hosted sandboxes, the chart creates static JuiceFS PVs and a static JuiceFS CSI config Secret in the LangSmith release namespace. Disabling client-config caching lets the bundled CSI resources keep Secret, pod exec/log, job/event, and PVC permissions scoped to that namespace while cluster-scoped permissions stay limited to discovery/watch operations.

This PR also adds checksum annotations for the chart-managed JuiceFS config path:

- `juicefs-csi-controller` and `juicefs-csi-node` roll when the chart-managed JuiceFS CSI config Secret data changes.
- `juicefs-csi-controller` and `juicefs-csi-node` roll when the chart-managed JuiceFS CSI driver ConfigMap changes.

These checksum annotations are intentionally not applied to `sandbox-host` or other LangSmith workloads.

Helm cannot hash live contents of an externally managed Secret, and it cannot directly refresh CSI-created JuiceFS mount pods. If an externally managed JuiceFS CSI Secret changes, delete affected JuiceFS mount pods so they are recreated with the updated Secret.

## Values Migration

This PR no longer reads the old top-level `juicefs-csi-driver` values block. Deployments that previously configured the subchart need to move those values under `config.sandboxes.juicefs.csi`:

- `juicefs-csi-driver.controller.annotations` -> `config.sandboxes.juicefs.csi.controller.annotations`
- `juicefs-csi-driver.node.annotations` -> `config.sandboxes.juicefs.csi.node.annotations`
- `juicefs-csi-driver.globalConfig.mountPodPatch` -> `config.sandboxes.juicefs.csi.mountPodPatch`

## Testing

- `git diff --check`
- `helm unittest charts/langsmith -f tests/sandboxes_test.yaml -f tests/sandbox_juicefs_csi_driver_test.yaml`
- `helm unittest charts/langsmith`
- `helm lint charts/langsmith`
- `helm template langsmith charts/langsmith -f charts/langsmith/tests/values/sandboxes-enabled.yaml`
- `helm template langsmith charts/langsmith -n default -f /tmp/self-hosted-ci-native-sandbox-values.yaml`
- `kubeconform -strict -ignore-missing-schemas /tmp/langsmith-inline-juicefs-render.yaml /tmp/langsmith-inline-juicefs-selfhosted-render.yaml`
- Targeted Kubernetes canary for the tightened CSI RBAC: namespace-scoped authorization checks plus JuiceFS PV/PVC mount read/write.
- End-to-end sandbox memory snapshot restore flow with a Python HTTP server kept alive across restore.


charts/langsmith/templates/sandboxes/juicefs-csi-driver/rbac.yaml
```diff
@@ -0,0 +1,293 @@
...
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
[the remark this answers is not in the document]
DanielKneipp: the role has been tightened, can you check again?

charts/langsmith/values.yaml
```diff
@@ -140,6 +140,31 @@ images:
     repository: "docker.io/langchain/sandbox-host"
     pullPolicy: IfNotPresent
     tag: ""
+  # -- JuiceFS CSI driver image. Only used when config.sandboxes.enabled is true.
+  juicefsCSIImage:
```
langchain-infra, line 144: only nit, do we need all these images?

langchain-infra (approved):

charts/langsmith/values.yaml
DanielKneipp: managed to remove the resizer

PR #822, opened 2026-07-22
feat(langsmith): wire SmithDB CM min/max replicas from Helm

ericdong-langchain:
New values under smithdb.clusterManager.services.{query,ingestion}:
- minReplicas (default 1)
- maxReplicas (helper to set max rep value)
- replicationThreshold (default 0)
-sliceReplicationThreshold (default 0)

Small-tier resource defaults
- disk path defaults

### DIFF
--- a/charts/langsmith/values.yaml
+++ b/charts/langsmith/values.yaml
@@ -1298,7 +1298,16 @@ smithdb:
       annotations: {}
       podSecurityContext: {}
       securityContext: {}
-      resources: {}
+      # -- Small-tier baseline. Override for larger deployments.
+      resources:
+        requests:
+          cpu: "4"
+          memory: "8Gi"
+          ephemeral-storage: "200Gi"
+        limits:
+          cpu: "4"
+          memory: "8Gi"
+          ephemeral-storage: "200Gi"
       probes:
         startupProbe:
           httpGet:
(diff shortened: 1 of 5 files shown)


charts/langsmith/values.yaml
```diff
@@ -1331,15 +1332,24 @@ smithdb:
       command:
         - "./smithdb"
         - "query"
-      extraEnv: []
+      extraEnv:
+        - name: SMITHDB_QUERY__VORTEX_CACHE__SEGMENT_CACHE__DISK_PATH
+          value: "/data/segment"
+        - name: SMITHDB_QUERY__VORTEX_CACHE__FILTER_CACHE__DISK_PATH
+          value: "/data/filter"
```
ben11211: Should be no longer required.

charts/langsmith/values.yaml
```diff
@@ -1529,13 +1575,30 @@ smithdb:
...
+      query:
+        minReplicas: 1
+        maxReplicas: ""
+        replicationThreshold: 0
+        sliceReplicationThreshold: 0
+      ingestion:
+        minReplicas: 1
+        maxReplicas: ""
+        replicationThreshold: 0
+        sliceReplicationThreshold: 0
```
ben11211: Is it possible to just pass replica count from query or ingestion directly into the clustermanager env rather than allowing these to not match potentially? If it should always be 0 for self host we should hardcode that I think.

ben11211 (approved):

PR #847, opened 2026-07-27
docs(langsmith): add stable-smithdb example values

asatish-langchain:
## Summary
- Adds `charts/langsmith/examples/stable_smithdb.yaml` documenting the "stable-smithdb" backend configuration: ClickHouse disabled, SmithDB as the sole ingestion/query backend.
- Sets only the flags that differ from LangSmith app code defaults; anything matching a default is intentionally omitted.
- Non-default rollout flags (`SMITHDB_FEEDBACK_MUTATION_*`, `SMITHDB_RULES_*_READ_ENABLED`, `RUN_RULES_SMITHDB_*_RETRIEVER_*`, `RUN_RULES_TWO_POINTER_BACKFILL_TENANTS`, `SMITHDB_QUERY_SERVICE_MAX_CALL_RECV_MSG_SIZE`) are injected via each LangSmith app service's `deployment.extraEnv` slot, since the chart does not natively expose them.

Follow-up PR will add the equivalent `stable_dual` example.

## Test plan
- [x] `helm template audit charts/langsmith -f <base-auth> -f charts/langsmith/examples/stable_smithdb.yaml` renders without error.
- [x] Rendered `-config` ConfigMap contains `SMITHDB_INGESTION_ENABLED=true`, `SMITHDB_QUERY_ENABLED=true`, `CLICKHOUSE_*_ENABLED=false`, `SMITHDB_CLUSTER_MANAGER_ENABLED=true`, and all `SMITHDB_*_SERVICE_URL` values.
- [x] `backend`, `queue`, and `ingestQueue` Deployments carry every FF env from the anchor list; `hostBackend` picks them up when enabled (verified with `--set hostBackend.enabled=true`).

### DIFF
--- a/charts/langsmith/examples/stable_smithdb.yaml
+++ b/charts/langsmith/examples/stable_smithdb.yaml
@@ -1,0 +1,78 @@
+# Reference values for a stable SmithDB-only LangSmith deployment.
+#
+# Backend layout:
+#   - ClickHouse fully disabled (no ingestion, no query, no in-chart StatefulSet).
+#   - SmithDB is the sole ingestion and query backend.
+#
+# Feature-flag intent mirrors the "stable-smithdb" column of the internal
+# rollout matrix. Only flags that differ from the LangSmith app code defaults
+# are set here; anything matching the code default is intentionally omitted.
+#
+# Placeholders that operators MUST replace:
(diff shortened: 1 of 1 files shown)


asatish-langchain:
Superseded — will land the flags as chart defaults gated on smithdb.enabled instead of adding an example file.

PR #851, opened 2026-07-27
test(langsmith): cover dual and smithdb-only backend modes

asatish-langchain:
Stacked on #848.

## Summary
- Extends `langsmith_smithdb_test.yaml` with an explicit **dual mode** assertion set on the existing `smithdb-enabled.yaml` fixture: confirms both `CLICKHOUSE_*_ENABLED` and `SMITHDB_*_ENABLED` render `"true"`, all new SmithDB flag envs fire, and `RUN_RULES_TWO_POINTER_BACKFILL_TENANTS` is present.
- Adds a new `tests/values/smithdb-only.yaml` fixture (`clickhouse.enabled: false` + SmithDB enabled).
- Two new tests exercise the smithdb-only fixture:
  1. ConfigMap: `CLICKHOUSE_*_ENABLED` render `"false"`, all SmithDB FF envs still fire (gate is `smithdb.enabled`, not the ingestion/query sub-flags).
  2. Workload gating: no ClickHouse StatefulSet, service, ConfigMap, secret, service-account, or migrations Job render.

No template or values.yaml changes — purely test additions.

Since the base PR already made all the SMITHDB-named FF envs gated on `smithdb.enabled` (which is true in both dual and smithdb-only), the runtime behavior for dual was already correct; this PR is about explicit test discrimination between the two modes.

## Test plan
- [x] `helm unittest charts/langsmith` — 120/120 pass (was 117 pre-PR; +3 new cases).

### DIFF
--- a/charts/langsmith/templates/config-map.yaml
+++ b/charts/langsmith/templates/config-map.yaml
@@ -56,6 +56,15 @@ data:
   {{- if .Values.smithdb.enabled }}
   SMITHDB_CLUSTER_MANAGER_ENABLED: "true"
   SMITHDB_CLUSTER_MANAGER_URL: {{ include "langsmith.smithdb.grpcServiceUrl" (dict "root" . "component" "clusterManager" "port" .Values.smithdb.clusterManager.service.port) | quote }}
+  SMITHDB_QUERY_SERVICE_MAX_CALL_RECV_MSG_SIZE: "1073741824"
+  SMITHDB_FEEDBACK_MUTATION_RPS_PER_KEY: "100"
+  SMITHDB_FEEDBACK_MUTATION_BURST_PER_KEY: "150"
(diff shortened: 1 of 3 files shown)


asatish-langchain:
Rebased on updated #848: dual + smithdb-only asserts now cover `RUN_RULES_REDIS_DEDUP_TENANTS` and `RUN_RULES_LIMIT_RRA_RECORDS_PER_BATCH_TENANTS` (both `["*"]`, unconditional). `RUN_RULES_TWO_POINTER_BACKFILL_TENANTS` asserts removed since that flag was dropped upstream per Natasha.

asatish-langchain:
Rebased on updated #848: _PCT asserts removed (redundant with wildcard tenants).

asatish-langchain:
Rebased on updated #848: PCT asserts restored in the dual and smithdb-only test cases.

PR #873, opened 2026-07-29
fix(langsmith): set TaskDB resource defaults

ben11211:
## Summary
- forward-port #872 to `main`
- give the chart-managed migration TaskDB guaranteed CPU and memory instead of BestEffort QoS

## Test plan
- [x] verify the rendered TaskDB StatefulSet includes all four resource defaults


langchain-infra (approved):
isnt guaranteed qos if limit = requests? but rest lgtm. jk i see you werent talking about guaranteed qos

PR #881, opened 2026-07-30
refactor(engine): drop the per-service JWT secrets from the co-host

baskaryan:
> **Draft: do not merge until the checklist below is complete.** Merging early breaks Engine dispatch — see Ordering.

## What

Removes the three per-service JWT secret env vars from the co-hosted deployment and from smith-go:

- `SMITH_BACKEND_SERVICE_JWT_SECRET` and `SMITH_GO_SERVICE_JWT_SECRET` from `langsmith.insights.extraEnv` (the co-host verifier side)
- `SMITH_GO_SERVICE_JWT_SECRET` from `langsmith.engine.smithGoEnv` (the signer side)

All three were set from `api_key_salt` — the same value `X_SERVICE_AUTH_JWT_SECRET` already carries on every pod via `conditionalEnvVars`. Three variables, one secret.

This is the change that actually collapses the configuration. langchain-ai/langchainplus#31926 only makes it *possible*, by teaching `combined_auth` to fall back to the shared secret and by selecting the service-key mode from the assistant id rather than from the secret being present.

## Ordering — the reason this is a draft

Against an image whose `combined_auth` predates the fallback, removing these leaves **no accepted key**, and every Engine dispatch returns 401. The failure is total and looks like a credentials problem rather than a config-ordering one.

- [ ] langchain-ai/langchainplus#31926 is merged
- [ ] the co-hosted image is rebuilt from a SHA containing it
- [ ] `self-hosted-ci`'s pinned `engineInsightsAgentImage.tag` is bumped to that image — it is marked `# nobump`, so it will not move on its own

I did not add a chart-side version guard: the image tags (`0.16.22rc1-43cf05f3`) are not valid semver, so `semverCompare` would either error or silently misjudge, and a guard that rejects legitimate tags is worse than a documented ordering.

## Test Plan

- [x] Renders with engine + sandboxes enabled: zero references to either per-service secret, 21 to the shared one.
- [x] Engine-disabled path renders unchanged.
- [x] `helm lint` clean across the CI values files.
- [ ] Verified on `self-hosted-ci` after the checklist above — not yet possible.


baskaryan:
Superseded by #893. This branch predated two `v16-stable` merges, so its diff had grown to a phantom 60 files against the moved base. #893 is the same three-line change against current `v16-stable`, with the safety argument re-verified: the dispatch switch now keys off `FORGE_AGENT_ASSISTANT_ID` rather than the secret's presence (langchainplus #31926, shipped in 0.16.27rc1).

PR #886, opened 2026-07-30
fix(langsmith): honor config.customCa in sandbox-host

langchain-infra:
`sandbox-host` was the only workload in the chart not wired to `config.customCa`, so an install that serves TLS from a private CA had no way to make the host trust it. Mounts the bundle and sets `SSL_CERT_FILE` via the same `langsmith.tlsVolumes` / `langsmith.tlsVolumeMounts` helpers every other component uses.

`LANGSMITH_SANDBOX_INTERNAL_ENDPOINT` is unchanged — still the in-cluster `platform-backend` Service.

Note for operators: `SSL_CERT_FILE` *replaces* the system trust store rather than adding to it, which is how the rest of the chart already behaves. The secret must therefore hold a full bundle, not just the private CA, or the host loses trust for anything else it talks to over TLS.

Stacked on `daniel/self-hosted-sandboxes-helm`.

## Test Plan

- [x] `helm unittest charts/langsmith` — 152 pass (2 new: CA mounted when configured, absent by default)
- [x] `helm lint`
- [x] Rendered with `config.customCa` set: `SSL_CERT_FILE` env, the `langsmith-custom-ca` volume, and a `subPath` mount that does not shadow `/etc/ssl/certs` or the `/certs` proxy-CA mount
- [ ] Confirm against a private-CA install that the host's calls to platform-backend and its snapshot pulls succeed

### DIFF
--- a/charts/langsmith/templates/sandboxes/sandbox-host-deployment.yaml
+++ b/charts/langsmith/templates/sandboxes/sandbox-host-deployment.yaml
@@ -1,6 +1,8 @@
 {{- if .Values.config.sandboxes.enabled }}
 {{- $namespace := .Values.namespace | default .Release.Namespace }}
 {{- $autoscaling := .Values.config.sandboxes.sandboxHost.autoscaling }}
+{{- $tlsVolumes := include "langsmith.tlsVolumes" . | fromYamlArray }}
+{{- $tlsVolumeMounts := include "langsmith.tlsVolumeMounts" . | fromYamlArray }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -94,6 +96,10 @@ spec:
                   optional: {{ .Values.config.disableSecretCreation }}
             - name: SANDBOX_HOST_PROXY_CA_DIR
               value: /certs
+            {{- if and .Values.config.customCa.secretName .Values.config.customCa.secretKey }}
+            - name: SSL_CERT_FILE
+              value: /etc/ssl/certs/custom-ca-certificates.crt
+            {{- end }}
             {{- if or (not (empty .Values.config.existingSecretName)) .Values.config.sandboxes.callbackSigningJwk }}
             - name: LANGSMITH_SANDBOX_CALLBACK_SIGNING_JWK
               valueFrom:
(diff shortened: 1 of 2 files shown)


langchain-infra:
Closing: this is all risk and no benefit.

`sandbox-host`'s egress proxy dials origins with `originTLSConfig = &tls.Config{MinVersion: tls.VersionTLS12}` (`smith-go/sandboxes/internal/proxy/proxy.go:148`) — `RootCAs` is nil, so Go uses the system cert pool, which honors `SSL_CERT_FILE`. Setting that env var on the host therefore replaces the trust store for **every origin the proxy dials on behalf of sandboxes**. A `config.customCa` secret holding only a private CA rather than a full bundle would break all sandbox outbound HTTPS (pypi, npm, github).

And the host gains nothing in return: it has no https targets of its own, and `LANGSMITH_SANDBOX_INTERNAL_ENDPOINT` is the in-cluster platform-backend Service over plain http.

If a private-CA install ever needs sandbox egress to trust an internal CA, that should be explicit — append the CA to `x509.SystemCertPool()` for `originTLSConfig` in smith-go, and/or inject it into the guest trust store — not a process-wide side effect of `config.customCa`.

langchain-infra:
Correction to the close comment: this would **not** break the MITM itself.

The serving leg is unaffected — the proxy signs leaf certs with the CA it loads explicitly from `tls.crt`/`tls.key` under `SANDBOX_HOST_PROXY_CA_DIR` (`egress_linux.go:47`), which is signing material, and the guest trusts that CA from its own injected store. Interception keeps working.

Only the **origin leg** is affected: `originTLSConfig` leaves `RootCAs` nil, so verifying the real upstream uses the system pool that `SSL_CERT_FILE` overrides. The failure mode is that interception succeeds and the proxy then fails to verify the genuine origin — so from inside a sandbox, every public HTTPS host fails with certificate errors while the proxy looks healthy.

Still the right call to close, since sandbox-host's only outbound call is plain http to the in-cluster platform-backend Service, so there was no upside to weigh against it.

langchain-infra:
Reopening: mounting the existing config.customCa is what's wanted here, no smith-go change.

PR #887, opened 2026-07-30
feat(langsmith): give sandbox egress an operator CA for origin verification

langchain-infra:
Chart side of langchain-ai/langchainplus#32452. **Needs a `sandbox-host` image containing that PR** — an older image ignores the variable, so the behavior is simply absent rather than broken.

Mounts `config.customCa` for `sandbox-host` and points `SANDBOX_HOST_EGRESS_ORIGIN_CA_FILE` at it, letting sandboxes reach internal HTTPS services signed by the operator's own CA.

**Why not `SSL_CERT_FILE`** — that's what I tried first in #886 and closed. Go's `x509.SystemCertPool()` honors it, so it *replaces* the trust store the egress proxy verifies origins with. The MITM keeps working (the proxy signs leaf certs from its own CA at `/certs`, unaffected), and the proxy then fails to verify genuine origins — so every public HTTPS host breaks inside sandboxes while the proxy looks healthy. The new variable appends to the system roots instead.

The CA is mounted at `/etc/langsmith/egress-ca/` rather than under `/etc/ssl/certs`, so it can't be mistaken for a trust-store override, and it doesn't collide with the `/certs` proxy-CA mount.

Nothing is needed inside the guest: all sandbox traffic is intercepted, so code in the VM only ever validates the proxy CA already in its trust store.

Also adds a section to `ENABLE-SANDBOXES.md` covering this, including that a configured-but-unreadable CA fails host startup rather than starting without it.

## Test Plan

- [x] `helm unittest charts/langsmith` — 152 pass (2 new: wired when `config.customCa` is set, absent by default, and asserts `SSL_CERT_FILE` is *not* set)
- [x] `helm lint`
- [x] Rendered with `config.customCa`: env, `subPath`-free secret mount at `/etc/langsmith/egress-ca`, no clash with `/certs`
- [ ] On a private-CA install with an image containing #32452: a sandbox reaches an internal HTTPS host, and public HTTPS still works in the same sandbox

### DIFF
--- a/charts/langsmith/docs/ENABLE-SANDBOXES.md
+++ b/charts/langsmith/docs/ENABLE-SANDBOXES.md
@@ -116,6 +116,21 @@ issuer and sandbox egress callbacks fail closed, since receivers fetch the verif
 Service auth between LangSmith and the sandbox runtime reuses `config.apiKeySalt`, the same secret
 the chart already uses for `X_SERVICE_AUTH_JWT_SECRET`. There is nothing extra to set.
 
+## Reaching internal HTTPS services from sandboxes
+
+Sandbox egress is intercepted by a MITM proxy on the host, so code inside a sandbox only ever
+validates the proxy's own CA — nothing in the guest needs your CA. The proxy is what verifies the
+real origin, and it does so against the system trust store.
+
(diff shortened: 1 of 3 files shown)


langchain-infra:
Not wanted — closing.

PR #909, opened 2026-08-04
chore(langsmith): default to the combined insights-engine image

baskaryan:
Enabling Engine currently takes two settings that have to agree:

```yaml
engine:
  enabled: true
images:
  engineInsightsAgentImage:
    repository: docker.io/langchain/langsmith-insights-engine   # or the render fails
```

because the default served only the `insights` graph. Those two disagreeing is what surfaced as a chart demanding an image customers had no way to pull.

## Change

Default `images.engineInsightsAgentImage.repository` to `langsmith-insights-engine`. `engine.enabled` is now sufficient on its own.

**The combined image is no larger:**

```
langsmith-clio              amd64 compressed:  397.0 MB   (amd64, arm64)
langsmith-insights-engine   amd64 compressed:  392.2 MB   (amd64, arm64)
```

5 MB smaller, same architectures — so insights-only installs pay nothing for a default that also carries the engine graph. Which of the two graphs actually runs is still `insights.enabled` / `engine.enabled`; the image only decides which are available.

## The validate guard stays

I had expected to delete it as unreachable. It isn't: an install that pinned `langsmith-clio` in its **own** values keeps that pin across the upgrade, and if it then enables Engine the guard is exactly what explains why. It only stops firing for installs that never overrode the default.

## Upgrade impact

Insights-only installs move to a different image on their next chart upgrade — same size, same graphs served, but a pod roll and a new image name in their registry. Anyone mirroring needs `langsmith-insights-engine` present first: that is #900 (merged) plus the Docker Hub repository now being public, so the prerequisites are in place.

## Test plan

- [x] New `engine_insights_agent_image_test.yaml`: the default is the combined image with insights alone, and enabling Engine needs no image change
- [x] Full chart suite: 291 passed (22 suites)
- [x] `helm lint` clean
- [x] `helm template` with `engine.enabled=true` and **no image override** now renders — that combination failed before
- [x] README regenerated with helm-docs v1.14.2

### DIFF
--- a/charts/langsmith/README.md
+++ b/charts/langsmith/README.md
@@ -560,7 +560,7 @@ The chart-managed frontend owns public API route rewrites for LangSmith services
 | images.clickhouseImage.repository | string | `"docker.io/clickhouse/clickhouse-server"` |  |
 | images.clickhouseImage.tag | string | `"25.12"` |  |
 | images.engineInsightsAgentImage.pullPolicy | string | `"IfNotPresent"` |  |
-| images.engineInsightsAgentImage.repository | string | `"docker.io/langchain/langsmith-clio"` |  |
+| images.engineInsightsAgentImage.repository | string | `"docker.io/langchain/langsmith-insights-engine"` |  |
 | images.engineInsightsAgentImage.tag | string | `"0.16.32rc1"` |  |
 | images.frontendImage.pullPolicy | string | `"IfNotPresent"` |  |
 | images.frontendImage.repository | string | `"docker.io/langchain/langsmith-frontend"` |  |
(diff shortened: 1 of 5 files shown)


langchain-infra (approved):

baskaryan:
Pushed a fix on top — the default flip had a bug that broke Insights on installs that don't enable Engine (Claude found it and wrote the fix).

`CLIO_ASSISTANT_ID: insights` was gated on `engine.enabled`, but the graph name follows the *image*, not Engine. The combined image exposes `insights` and `engine`; only the retired clio image had the `agent` graph that `lc_config` defaults to (`lc_config/lc_config/settings.py:1589`). So with `engine.enabled=false` — the default, and the only option today on GCP, Azure and EU — this PR rendered the combined image with no `CLIO_ASSISTANT_ID`, pointing smith-backend at a graph that doesn't exist.

Confirmed by rendering: engine off gave `image: langsmith-insights-engine` and no `CLIO_ASSISTANT_ID`. The override now sits directly under `insights.enabled` in both the backend and queue deployments.

Three tests added, and I verified they fail against the pre-fix templates rather than passing vacuously.

Checked the rest of the engine-off path while I was in there, and it's clean: `X_SERVICE_AUTH_JWT_SECRET` comes from `conditionalEnvVars` (derived from `api_key_salt`) so it reaches both sides regardless of `engine.enabled`; `combined_auth` reads its config lazily; `get_agent` defers its heavy imports below an early return, so the engine graph registers at startup with no Engine config; and the two images are the same size (392 vs 393 MB).

PR #919, opened 2026-08-05
chore(langsmith): bump chart to 0.16.0-rc.30 / appVersion 0.16.34

baskaryan:
Cuts a chart release carrying everything that has landed on `v16-stable` since `rc.29`, and moves onto the `0.16.34` images.

### What this releases

**Chart changes** — both landed after the `rc.29` bump (#911) and are currently unpublished:

- #910 — drop the `langsmith-clio` image (and fix the marketplace scripts, which published clio and not the combined image)
- #914 — opt orgs out of Engine when Engine is not deployed

**App changes** — `appVersion` moves `0.16.33rc1` → `0.16.34`, which picks up:

- #33034 — self-hosted Engine license entitlement gate
- #33035 — hide Engine settings where Engine is not deployed

Both merged at 09:36/09:38 UTC, before the `0.16.34` cut (#33021) at 16:25 UTC, and I confirmed `EngineEnabledClaim` is present in the tree at that bump commit — so the gate really is in these images.

### Verification

- All seven images the chart references confirmed present at `0.16.34` on Docker Hub.
- 299 tests pass across 22 suites; `helm lint` clean; README regenerated.
- No stale `0.16.33rc1` tags remain in `values.yaml`.

### Before merging — the entitlement gate is live in these images

Any cluster that upgrades to this chart with `SMITH_INTELLIGENCE_ENDPOINT` set (i.e. `engine.enabled=true`) will **refuse to start** unless its license carries `engine_enabled`.

Our internal clusters are already covered: the entitlement was added to the `eks-smith-langchain` customer, I verified beacon mints it (`engine_enabled: true` alongside `sandbox_enabled`), and all three clusters had their cached license refreshed. Any other install enabling Engine needs the same treatment first.

### Note on versioning

`0.16.34` dropped the `rc` suffix, which by convention means the v16 app train has gone GA. I've kept the chart on the `rc.N` series as asked, but if the chart should follow the app to a GA `0.16.0`, that's a release-management call worth making deliberately rather than inheriting from this PR.

### DIFF
--- a/charts/langsmith/Chart.yaml
+++ b/charts/langsmith/Chart.yaml
@@ -5,5 +5,5 @@ maintainers:
     email: ankush@langchain.dev
 description: Helm chart to deploy the langsmith application and all services it depends on.
 type: application
-version: 0.16.0-rc.29
-appVersion: "0.16.33rc1"
+version: 0.16.0-rc.30
+appVersion: "0.16.34"
--- a/charts/langsmith/README.md
+++ b/charts/langsmith/README.md
@@ -1,6 +1,6 @@
 # langsmith
 
-![Version: 0.16.0-rc.29](https://img.shields.io/badge/Version-0.16.0--rc.29-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.16.33rc1](https://img.shields.io/badge/AppVersion-0.16.33rc1-informational?style=flat-square)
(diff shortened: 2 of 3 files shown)


baskaryan:
Superseded by #913, which cut chart `0.16.0` GA and already carries everything this PR was for:

- #910 (clio dropped from the mirror script) — in
- #914 (Engine kill-switch when Engine is not deployed) — in
- image tags on `0.16.34` — in

I opened this against `rc.29` before #913 landed, so it now conflicts and would only re-litigate the version number. Closing rather than resolving.

The one item from this PR's description still worth carrying forward: `0.16.34` images contain the Engine license entitlement gate (#33034), so any install upgrading with `engine.enabled=true` needs `engine_enabled` on its license first. Our internal clusters are already covered.

PR #920, opened 2026-08-06
fix(langsmith): mount /data for the SmithDB query disk cache

ericdong-langchain:
## Summary
The SmithDB `query` deployment enables the Vortex disk cache but never gets a writable `/data` volume, so the cache silently never initialises.

## Changes
- `values.yaml`: give `smithdb.query.deployment` an emptyDir `local-ssd-storage` mounted at `/data`, `sizeLimit` matching the 200Gi ephemeral-storage limit, mirroring `ingestion`.

## Test plan

- [x] `helm unittest charts/langsmith -f 'tests/langsmith_smithdb_test.yaml'` — 39 passed, including the new case
- [x] `helm lint charts/langsmith` clean

### DIFF
--- a/charts/langsmith/values.yaml
+++ b/charts/langsmith/values.yaml
@@ -1485,8 +1485,16 @@ smithdb:
       tolerations: []
       affinity: {}
       topologySpreadConstraints: []
-      volumes: []
-      volumeMounts: []
+      # The Vortex disk cache writes to /data. Without a writable mount the
+      # cache silently degrades to memory-only and every miss goes to object
+      # storage. Keep sizeLimit aligned with resources.limits.ephemeral-storage.
+      volumes:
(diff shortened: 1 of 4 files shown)


charts/langsmith/values.yaml
```diff
@@ -1485,8 +1485,16 @@ smithdb:
       tolerations: []
       affinity: {}
       topologySpreadConstraints: []
-      volumes: []
-      volumeMounts: []
+      # The Vortex disk cache writes to /data. Without a writable mount the
+      # cache silently degrades to memory-only and every miss goes to object
+      # storage. Keep sizeLimit aligned with resources.limits.ephemeral-storage.
```
ben11211: I think this comment can be removed otherwise LGTM!

ben11211 (approved):

PR #945, opened 2026-08-13
feat(langsmith): add config.settings.webhookRequestTimeoutSec

mcschan:
## What

Adds `config.settings.webhookRequestTimeoutSec`, defaulted to `"5"`, and templates it into `langsmith.commonEnv` as `WEBHOOK_REQUEST_TIMEOUT_SEC`.

## Why

Automation rule and prompt webhooks had a hardcoded 5 second timeout in the backend. [LSO-810](https://linear.app/langchain/issue/LSO-810/request-increase-webhook-timeouts-from-5s-to-10s) asks for 10s — customer endpoints often need longer than 5s, and today they work around it by returning `200` immediately and processing asynchronously.

The companion `langchainplus` change makes the timeout configurable via `WEBHOOK_REQUEST_TIMEOUT_SEC` and defaults it to **10s on LangSmith Cloud, 5s on self-hosted**. Self-hosted stays at 5s deliberately: a longer timeout holds a rules worker slot longer, and we can neither observe nor quickly roll back those deployments.

This PR gives operators a first-class knob so raising it doesn't require hand-editing `commonEnv`.

## Notes

- Templated into `langsmith.commonEnv` rather than a per-service `extraEnv` because **automation rules fire from the queue worker, not the API backend** — `backend.deployment.extraEnv` alone would silently have no effect.
- Because the chart emits the variable unconditionally, Helm installs always set it explicitly and the chart is the single source of truth. The backend's own self-hosted default (also 5) still covers non-Helm self-hosted installs, so the two agree.
- Safe to merge before the backend ships: `SharedSettings` is configured `extra="ignore"`, so an older image treats the variable as a no-op.
- Chart version bumped `0.17.0-rc.6` → `0.17.0-rc.7` for `ct lint`'s version-increment check. `README.md` rows added by hand in helm-docs format (helm-docs isn't installed locally); worth a skim.

## Test plan

- [ ] CI `ct lint` and `helm unittest` pass
- [ ] `helm template` shows `WEBHOOK_REQUEST_TIMEOUT_SEC: "5"` on the queue and backend deployments
- [ ] Setting `config.settings.webhookRequestTimeoutSec: "10"` renders `"10"`


mcschan:
Closing as this is not necessary per @samnoyes.

PR #952, opened 2026-08-17
fix(langsmith): default service listeners to dual-stack

rashiq:
## Summary

Removes the chart-managed `BIND_ADDRESS`. The `0.17.10rc1` image now handles dual-stack binding by default.

## Test Plan

- [x] Verified all services on IPv4 and IPv6-only kind clusters without IPv6-specific chart values.


charts/langsmith/values.yaml
```diff
@@ -1267,7 +1267,8 @@ config:
 aceBackend:
   name: "ace-backend"
   containerPort: 1987
-  bindAddress: "0.0.0.0"
+  # -- Address the ace-backend listener binds to. Empty lets the service bind every interface, which answers on IPv4 and IPv6. Set a single address only to restrict the listener.
+  bindAddress: ""
```
langchain-infra: can we just remove this arg? and then every service handles binding automatically?

rashiq: good catch!

rashiq: actually would need to fix deployment.yaml before I can remove this arg

rashiq:
`langster-safety` required the gitignore changes

langchain-infra (approved):

PR #1006, opened 2026-09-03
feat(langsmith): support separate sandbox clusters

DanielKneipp:
## Summary

- Preserve `sameCluster` as the default sandbox topology while adding an opt-in `separateCluster` mode that keeps sandbox APIs and workers in the LangSmith release but omits embedded runtime resources.
- Add a standalone `langsmith-sandbox` chart for `sandbox-host`, lease RBAC, JuiceFS formatting and mounts, proxy CA management, workload identity, and optional authenticated host ingress.
- Route control-plane and worker traffic through `SANDBOX_HOST_INGRESS_ENDPOINT` when configured; direct node-IP routing on TCP 19190 remains the default.
- Require the standalone runtime to consume a minimal existing Secret instead of copying the full LangSmith application Secret.

Companion PRs:
- AWS Terraform example: https://github.com/langchain-ai/terraform/pull/231
- Documentation: https://github.com/langchain-ai/docs/pull/5856

## Chart convention alignment

Reviewing the new chart against its siblings turned up three deviations, all fixed:

- **Missing `.chart` helper.** `langsmith-sandbox` inlined `printf` for the `helm.sh/chart` label, dropping `trunc 63`. A chart version carrying build metadata would produce a label over the 63-character limit and be rejected by the API server.
- **Dropped `common*` plumbing.** The new chart did not consume `commonVolumes`, `commonVolumeMounts`, or `commonInitContainers`, which the `langsmith` chart's sandbox-host does. Migrating `sameCluster` to `separateCluster` would have silently dropped that configuration.
- **Missing `commonDnsConfig`.** Absent from sandbox-host in both charts. It matters more in the new topology, where sandbox-host resolves an external hostname to reach the platform API.

The last two are applied to the `langsmith` chart's sandbox-host as well, closing the inherited gap rather than only fixing the copy.

## Test Plan

- [x] All 337 `langsmith` Helm unit tests across 27 suites.
- [x] All 21 `langsmith-sandbox` Helm unit tests.
- [x] Regression tests added for label truncation, `common*` plumbing, `dnsConfig` application, and the sandbox-host ServiceAccount name.
- [x] Lint both charts and package the standalone chart; `ct lint` passes with version-increment checking enabled.
- [x] Validate same-cluster, separate-cluster, and host-ingress manifests with kubeconform.
- [x] Verified the runtime resource set moves out of the `langsmith` chart and reappears in `langsmith-sandbox` one-for-one, and that the control-plane env is identical between modes — only runtime placement changes, not API behavior.
- [x] CI green (11 checks), including `lint-test (langsmith)` and `lint-test (langsmith-sandbox)`. Both run a real `ct install` against a kind cluster, so the charts install rather than merely render.
- [ ] Deploy both releases across routed test clusters and complete a sandbox create/execute/snapshot/restore smoke test.

Generated with [Devin](https://devin.ai)


DanielKneipp:
CI note: the `langsmith` matrix failed twice before chart installation while provisioning the workflow's floating Bitnami `redis:latest` TLS fixture. The Redis pod repeatedly failed its own readiness/liveness probes and timed out; chart version checks, lint, all 329 LangSmith unit tests, the new chart's 12 tests, and the `langsmith-sandbox` install job passed. This appears isolated to the shared CI fixture rather than the PR changes.

ramon-langchain (approved):

PR #1030, opened 2026-09-14
Add opt-in SmithDB recovery strategy

thobe:
## Summary

- add `smithdb.recoveryStrategy.enabled` as an opt-in setting, defaulting to `false`
- explicitly configure SmithDB recovery as disabled until the customer opts in
- add `smithdb.recoveryStrategy.retentionWindow`, defaulting to `12h`
- render the chart-owned settings directly in the compaction container environment
- validate the new settings and keep the chart README focused on values reference

## Rollout

1. Upgrade the full LangSmith installation to V17 while `smithdb.recoveryStrategy.enabled=false`.
2. Wait for all SmithDB query, compaction coordinator, and compaction worker pods to complete their V17 rollout.
3. In a separate Helm upgrade, set `smithdb.recoveryStrategy.enabled=true`.

Once enabled on V17, it must not be disabled again, and the installation must not be downgraded to V16. The customer-facing sequence belongs in the SmithDB upgrade guide in the docs repository rather than this chart README.

## Validation

- Helm lint
- default V17 render verifies recovery is explicitly disabled
- opt-in render verifies the configured retention window
- invalid empty retention window is rejected
- changed YAML parses successfully
- `git diff --check`


charts/langsmith/templates/smithdb/compaction-deployment.yaml
```diff
@@ -1,6 +1,12 @@
 {{- if .Values.smithdb.enabled }}
 {{- $resources := include "langsmith.smithdb.resources" (dict "root" . "component" "compaction") | fromYaml -}}
-{{- $envVars := concat (include "langsmith.smithdb.componentEnv" (dict "root" . "service" "COMPACTION" "displayName" "smithdb-compaction") | fromYamlArray) .Values.smithdb.compaction.deployment.extraEnv -}}
+{{- $recoveryStrategyEnv := list
+      (dict "name" "SMITHDB_COMPACTION__LARGE_FIELD_DELETE__ENABLE_LARGE_FIELD_VERSIONED_PATHS" "value" (toString .Values.smithdb.recoveryStrategy.enabled)) -}}
+{{- if .Values.smithdb.recoveryStrategy.enabled -}}
+{{- $recoveryStrategyEnv = append $recoveryStrategyEnv
+      (dict "name" "SMITHDB_COMPACTION__RECOVERY_WINDOW_SLA" "value" .Values.smithdb.recoveryStrategy.retentionWindow) -}}
+{{- end -}}
```
ben11211: nit: Would recommend moving these into env block for chart clarity.  (matching sequenced backfill pattern)

thobe: Addressed in a696ba28 by rendering these directly in the container env block alongside the sequenced-backfill setting.

langchain-infra (approved):

ericdong-langchain (approved):

